PR #28, opened 2020-02-13
Introduce ValidatorFn and BlackList implementation

ndeloof:
- Remove `UnsupportedProperties` which is swarm-specific
- Introduce a flexible `ValidatorFn` option for implementation to check supported features
- Offer a `Blacklist` helper factory to set ValidatorFn based on a JSONPath blacklist


bruno-oliveira:
@ndeloof this one looks good to me :+1:

PR #40, opened 2020-04-10
Cleanup unused types

### LINKED ISSUE #11: remove UnsupportedProperties / DeprecatedProperties / ForbiddenProperties
compose-go do define `types.UnsupportedProperties`, `types.DeprecatedProperties` and `types.ForbiddenProperties` which are actually only used by https://github.com/docker/cli and should not be part of reference library as "supported properties" is implementation specific

bruno-oliveira:
Cleans up unused, deprecated and forbidden properties. Fixes #11

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -106,10 +106,6 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 			return nil, errors.Errorf("version mismatched between two composefiles : %v and %v", configDetails.Version, version)
 		}
 
-		if err := validateForbidden(configDict); err != nil {
-			return nil, err
-		}
-
 		if !opts.SkipInterpolation {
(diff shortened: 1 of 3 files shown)


ndeloof:
Thansk for contributing. Isn't this the same objective as https://github.com/compose-spec/compose-go/pull/28 ?

bruno-oliveira:
Seems like it is, at least partially, but, should it done together? Or is this one a separate issue that addresses only part of the problem in #28? So, if this is done in isolation, can #28 be picked up from there?

ndeloof:
Happy to get your review on #28 then we could merged it

bruno-oliveira:
It looks good to me!

bruno-oliveira:
Since this one requires a single approval maybe it can be merged and then 28?

PR #45, opened 2020-05-14
Introduce `Project` as named, merge result for a set of config files

ndeloof:
This is code I had to write to implement a Compose compliant command line.

- introduce Project for a set of merged compose files (to replace use of Config)
- load Compose project from CLI options
- manage project Name and WorkingDir
- set Compose model defaults

Getting this hosted in compose-go make this code reusable in another context, so we don't have to duplicate code


cli/options.go
```diff
@@ -0,0 +1,161 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/compose-spec/compose-go/loader"
+	"github.com/compose-spec/compose-go/types"
+	"github.com/sirupsen/logrus"
+)
+
+// ProjectOptions group the command line options recommended for a Compose implementation
```
chris-crone, line 31:
```suggestion
// ProjectOptions groups the command line options recommended for a Compose implementation
```

cli/options.go
```diff
@@ -0,0 +1,161 @@
...
+type ProjectOptions struct {
+	ConfigPaths []string
+	Name        string
+}
+
+// SupportedFilenames define the supported Compose file names for auto-discovery (in preference order)
```
chris-crone, line 37:
```suggestion
// SupportedFilenames defines the supported Compose file names for auto-discovery (in order of preference)
```

cli/options.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+
+	return loader.Load(types.ConfigDetails{
+		ConfigFiles: configs,
+		Environment: environment(),
+	}, func(opts *loader.Options) {
+		if options.Name != "" {
+			opts.Name = options.Name
+		} else if nameFromEnv, ok := os.LookupEnv("COMPOSE_PROJECT_NAME"); ok {
```
chris-crone, line 58: We should make `COMPOSE_PROJECT_NAME` a const

cli/options.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+			if _, err := os.Stat(f); err != nil {
+				return nil, err
+			}
+			paths = append(paths, f)
+		}
+		return paths, nil
+	}
+
+	sep := os.Getenv("COMPOSE_FILE_SEPARATOR")
```
chris-crone, line 89: We should make `COMPOSE_FILE_SEPARATOR` a const

cli/options.go
```diff
@@ -0,0 +1,161 @@
...
+	if sep == "" {
+		sep = string(os.PathListSeparator)
+	}
+	f := os.Getenv("COMPOSE_FILE")
```
chris-crone, line 93: We should make `COMPOSE_FILE` a const

cli/options.go
```diff
@@ -0,0 +1,161 @@
...
+			winner := candidates[0]
+			if len(candidates) > 1 {
+				logrus.Warnf("Found multiple config files with supported names: %s", strings.Join(candidates, ", "))
+				logrus.Warnf("Using %s\n", winner)
+			}
+			return []string{winner}, nil
+		}
+		parent := filepath.Dir(pwd)
+		if parent == pwd {
+			return nil, fmt.Errorf("can't find a suitable configuration file in this directory or any parent. Are you in the right directory?")
```
chris-crone, line 116: Since this is a library, we should make the errors easy to handle for consumers. I'd recommend adding an errdefs package and doing something like `errors.Wrap(errdefs.ErrNotFound, "unable to find config file")`

It's also best to avoid giving user advice in the error itself. Stick to what's wrong.

cli/testdata/simple/compose.yaml
```diff
@@ -0,0 +1,4 @@
+version: "3"
+services:
+  simple:
+    image: nginx
```
chris-crone, line 4: nit: `\n`

cli/options.go
```diff
@@ -0,0 +1,169 @@
...
+	}, func(opts *loader.Options) {
+		if options.Name != "" {
+			opts.Name = options.Name
+		} else if nameFromEnv, ok := os.LookupEnv(ComposeProjectName); ok {
+			opts.Name = nameFromEnv
+		}
+	})
+}
+
+// getConfigPathsFromOptions retrieve the config files for project based on project options
```
rumpl, line 72:
```suggestion
// getConfigPathsFromOptions retrieves the config files for a project based on project options
```

cli/options.go
```diff
@@ -0,0 +1,169 @@
...
+	files := []types.ConfigFile{}
+	for _, f := range configPaths {
+		var (
+			b   []byte
+			err error
+		)
+		if f == "-" {
+			b, err = ioutil.ReadAll(os.Stdin)
+		} else {
+			if _, err := os.Stat(f); err != nil {
```
rumpl, line 140: Do we really need the stat here? `ReadFile` will return an error if it can't read it

compatibility/checker.go
```diff
@@ -0,0 +1,380 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compatibility
+
+import "github.com/compose-spec/compose-go/types"
+
+type Checker interface {
```
chris-crone, line 21: I'm a little worried about manually maintaining all of these.

Would it make more sense to only allow checking a service, network, volume (for example) and let the consumer of the library dive down into the exact specifics?

ndeloof: The idea is to offer a full implementation that will reject everything, and compose implementation to "opt-in" for a whitelist / override a few checkers when they need finer control (to check supported values). Doing so, we can introduce new attributes in compose spec, and compose implementations won't have to make changes for those to be ignored (with a warning), just update compose-go. Look at the test case for an illustration sample.

About maintenance, this initial implementation is huge, but subsequent maintenance will just be about introducing a new `Check` method when we introduce a new attribute.

ndeloof: That being said I'd like we can generate such code, as well as the compose-go types, from json schema. Do you know any library doing so we could get inspired by?

chris-crone: Looking at the tests, consuming this seems to be quite elegant. The only thing that would make it better is consts instead of free strings.

If we have a checklist on new additions to the spect that includes adding the required check code then this is fine.

I don't know of any libraries or tools autogenerating from JSON schema. Maybe @thaJeztah or @silvin-lubecki will though?

ndeloof:
Need to be updated to cover https://github.com/compose-spec/compose-spec/pull/78

compatibility/build.go
```diff
@@ -0,0 +1,77 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package compatibility
+
+import "github.com/compose-spec/compose-go/types"
+
+func (c *WhiteList) CheckBuild(service *types.ServiceConfig) bool {
```
thaJeztah, line 21: Just catching my eye; should we `s/WhiteList/AllowList/`?

compatibility/checker.go
thaJeztah: The JSON schema itself would allow for (e.g.) `"additionalProperties": false` to reject anything that's not defined, but that would mean that each driver would maintain a copy of the schema, and only add the properties they support. Perhaps that would be a two-step process;

1. Use the "upstream" schema to see if the file is well-formed according to the compose spec
2. Use the "implementation" JSON schema to invalidate / warn / error on options that are not supported by the specific implementation


I'm still a bit ... on the huge "interface"; are we using an interface so that implementations are _enforced_ to check for each property?

MAINTENANCE.md
```diff
@@ -0,0 +1,9 @@
+# Maintenance
+
+Compose-go library has to be kep tup-to-date with approved changes in [Compose-spec](https://github.com/compose-spec/compose-spec).
```
justincormack, line 3: typo "kep tup"

MAINTENANCE.md
```diff
@@ -0,0 +1,9 @@
...
+This typically require, as we define new attributes to be added to the spec
+
+1. Update `shcema` to keep
```
thaJeztah, line 6:
```suggestion
1. Update `schema` to keep   
```

justincormack, line 6: typo

compatibility/build.go
ndeloof: Not being a native english speacker I can't really tell, but I know this approach to opt-in for a set of features as "WhiteList"

```golang
mychecker := WhiteList{
			Supported: []string{
				"services.network_mode",
				"services.privileged",
				"services.networks",
			}
```

cli/options.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+// ProjectOptions groups the command line options recommended for a Compose implementation
+type ProjectOptions struct {
+	ConfigPaths []string
+	Name        string
+}
+
+// SupportedFilenames defines the supported Compose file names for auto-discovery (in order of preference)
+var SupportedFilenames = []string{"compose.yaml", "compose.yml", "docker-compose.yml", "docker-compose.yaml"}
```
justincormack, line 40: "Supported" is weird, maybe"Discovery" or "Default"?

ndeloof:
@thaJeztah I'm not using an interface here with the expectation anyone would write a custom implementation without using WhiteList as a base. I wish I could create an `AbstractChecker`.
Is there a better golang-compliant approach to define default methods implementation but give a chance for implementors to override?

compatibility/checker.go
```diff
@@ -0,0 +1,404 @@
...
+		project.Configs[i] = config
+	}
+
+	for i, secret := range project.Secrets {
+		CheckSecretsConfig(&secret, c)
+		project.Secrets[i] = secret
+	}
+}
+
+func CheckServiceConfig(service *types.ServiceConfig, c Checker) {
```
justincormack, line 204: why arent these methods of Checker?

ndeloof: Checker defines the per-attribute validation rule.
Those methods iterate on all attributes to call the adequate Checker

compatibility/build.go
thaJeztah: Yes, it's been commonly referred to as a "whitelist"; recent developments have raised questions wether "blacklist"/"whitelist" is a good term for these, or if "blocklist" / "denylist" / "allowlist" are more suitable.

compatibility/checker.go
ndeloof: With an Object-Oriented approach I would actually have set `CheckService` as part of the Default implementation, which would invoke sub Checkers, and give an option for implementors to override and skip/re-implement the whole service checking on it's own. But this is Golang

MAINTENANCE.md
```diff
@@ -0,0 +1,9 @@
+# Maintenance
+
+Compose-go library has to be kept up-to-date with approved changes in [Compose-spec](https://github.com/compose-spec/compose-spec).
```
chris-crone, line 3:
```suggestion
The compose-go library has to be kept up-to-date with approved changes in the [Compose spec](https://github.com/compose-spec/compose-spec).
```

compatibility/allowlist.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/pkg/errors"
+)
+
+// AllowList implement Checker interface by rejecting all attributes but those listed as "supported".
+type AllowList struct {
+	Supported []string
+	errors    []error
+}
+
+func (c *AllowList) Errors() []error {
```
chris-crone, line 32: Public function missing a comment, we should add a linter to pick things like this up (not in this PR).

```suggestion
// Errors returns the list of errors encountered when checking against the allow list
func (c *AllowList) Errors() []error {
```

compatibility/allowlist_test.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"testing"
+
+	"github.com/compose-spec/compose-go/errdefs"
+	"github.com/compose-spec/compose-go/loader"
+	"github.com/compose-spec/compose-go/types"
+	"gotest.tools/v3/assert"
+)
+
+func TestWhiteList(t *testing.T) {
```
chris-crone, line 28:
```suggestion
func TestAllowList(t *testing.T) {
```

compatibility/deploy.go
```diff
@@ -0,0 +1,224 @@
...
+}
+func (c *AllowList) CheckDeployLabels(config *types.DeployConfig) {
+	if !c.supported("services.deploy.labels") && len(config.Labels) != 0 {
+		config.Labels = nil
+		c.error("services.deploy.labels")
+	}
+}
+
+const (
+	UpdateConfigUpdate   = "update_config"
```
chris-crone, line 54: Where are these consts used?

compatibility/deploy.go
```diff
@@ -0,0 +1,224 @@
...
+func (c *AllowList) CheckUpdateConfigOrder(s string, config *types.UpdateConfig) {
+	k := fmt.Sprintf("services.deploy.%s.order", s)
+	if !c.supported(k) && config.Order != "" {
+		config.Order = ""
+		c.error(k)
+	}
+}
+
+const (
+	ResourceLimits       = "limits"
```
chris-crone, line 120: Where are these consts used?

compatibility/allowlist.go
ndeloof: as this is an interface method, this seems to me a very bad practice to require a comment (until there's some implementation-specific things to say)

compatibility/deploy.go
ndeloof: CheckUpdateConfigDelay(UpdateConfigUpdate, service.Deploy.UpdateConfig)
CheckUpdateConfigDelay(UpdateConfigRollback, service.Deploy.RollbackConfig)

compatibility/deploy.go
ndeloof: CheckDeployResourcesNanoCPUs(ResourceLimits, service.Deploy.Resources.Limits)
CheckDeployResourcesNanoCPUs(ResourceReservations, service.Deploy.Resources.Limits)

Golang do not support method override 😥

compatibility/allowlist.go
chris-crone: That makes sense to me. The linter might actually take that into account.

PR #55, opened 2020-06-25
introduce WorkingDir and Environment in ProjectOptions

ndeloof:
Move definition of `WorkingDir` and `Environment` to `cli.ProjectOptions` so that `Loader` do not handle filepath default magic and is not responsible to CLI specific things


cli/options.go
```diff
@@ -18,6 +18,7 @@ package cli
 
 import (
 	"fmt"
+	"github.com/joho/godotenv"
```
rumpl, line 21: This one should go in the group on the bottom

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/docker/go-units v0.4.0
 	github.com/google/go-cmp v0.5.0
 	github.com/imdario/mergo v0.3.8
+	github.com/joho/godotenv v1.3.0 // indirect
```
rumpl, line 10: I wonder why is this marked as indirect ?

PR #58, opened 2020-07-07
Introduce ErrIncompatible

ndeloof:
Define a higher-severity error type for compatibility checker to report incompatible attributes or values that make the compose file impossible to be used by Compose implementation. ErrUnsupported will result into ignored attributes.

### DIFF
--- a/errdefs/errors.go
+++ b/errdefs/errors.go
@@ -27,6 +27,9 @@ var (
 
 	// ErrUnsupported is returned when a compose project uses an unsupported attribute
 	ErrUnsupported = errors.New("unsupported attribute")
+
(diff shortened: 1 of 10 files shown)


errdefs/errors.go
```diff
@@ -27,6 +27,9 @@ var (
 
 	// ErrUnsupported is returned when a compose project uses an unsupported attribute
 	ErrUnsupported = errors.New("unsupported attribute")
+
+	// ErrIncompatible is returned when a compose project uses an incompatible attribute
+	ErrIncompatible = errors.New("incompatible attribute")
```
chris-crone, line 32: How does this differ to unsupported? Is it that there are a combination of features being used that does not work?

PR #60, opened 2020-07-13
Discard a service's env_file using project options.

ezodude:
The `discardEnvFiles` `loader.Load` option is not visible to developers wanting to create a project using `cli.ProjectFromOptions(&projectOptions)`.

This PR,

- Adds the `loadOptions` property to the `ProjectOptions` struct.

- Enables discarding the `env_file` using the new `WithDiscardEnvFile` method.

- `WithDiscardEnvFile` adds `loader.WithDiscardEnvFiles` func to `ProjectOptions.loadOptions`.

- Then the load options are passed to `loader.load` in the `ProjectFromOptions` func.


ndeloof:
note: `loader.Options#discardEnvFiles` was introduced by https://github.com/docker/cli/commit/821f5ecf53c87a1f391d97c358f08c8d344cda8c

Makes perfect sense to me to introduce `LoadOptions` in `ProjectOptions` (we should define a func type for it)

ndeloof:
I'm sorry I didn't reviewed this before, could you please rebase then I can merge ?

cli/options.go
```diff
@@ -117,6 +118,18 @@ func WithDotEnv(o *ProjectOptions) error {
 	return nil
 }
 
+// WithDiscardEnvFiles sets discards the `env_file` section after resolving to
+// the `environment` section
+func (o ProjectOptions) WithDiscardEnvFile() (ProjectOptions, error) {
```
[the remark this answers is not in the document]
ezodude: Hey, thanks for the review 🙂

So, `WithDiscardEnvFile` doesn't require extra info to create a `ProjectOptionsFn`, it's the actual function (very similar to `WithOsEnv`). So, I'm updating it accordingly.

ezodude:
@ndeloof all done.

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
ndeloof, line 1: seems we're missing a rule in `.gitignore` to exclude golang configuration files :)

ezodude: Oops will add and ignore the .idea file.

PR #61, opened 2020-07-15
Prefer ProjectOption over *ProjectOption for homogeneity

ndeloof:
Make code homogeneous regarding references to `ProjectOption` (pointer vs value)

Introduce ProjectOptions constructor with functional parameters
This allows Compose implementations to use:
```
opts, err := NewProjectOptions( []string{ "mycompose.yaml"}, 
   WithOsEnv, WithDotEnv, WithEnv(variables), WithName("test") 
)
ProjectFromOptions( opts )
```

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -37,53 +37,84 @@ type ProjectOptions struct {
 	Name        string
 	WorkingDir  string
 	ConfigPaths []string
-	Environment []string
+	Environment map[string]string
 }
 
-// WithOsEnv imports environment variables from OS until those have been overridden by ProjectOptions.Environment
-func (o ProjectOptions) WithOsEnv() ProjectOptions {
-	env := getAsEqualsMap(o.Environment)
-	for k, v := range getAsEqualsMap(os.Environ()) {
-		if _, ok := env[k]; !ok {
-			env[k] = v
+type ProjectOptionsFn func(*ProjectOptions) error
+
+// NewProjectOptions creates ProjectOptions
+func NewProjectOptions(configs []string, opts ...ProjectOptionsFn) (*ProjectOptions, error) {
+	options := &ProjectOptions{
+		ConfigPaths: configs,
+		Environment: map[string]string{},
+	}
+	for _, o := range opts {
+		err := o(options)
+		if err != nil {
+			return nil, err
 		}
 	}
+	return options, nil
+}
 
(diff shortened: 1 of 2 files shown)


cli/options.go
```diff
@@ -37,53 +37,84 @@ type ProjectOptions struct {
 	Name        string
 	WorkingDir  string
 	ConfigPaths []string
-	Environment []string
+	Environment map[string]string
```
thaJeztah, line 40: Haven't looked closely at how it's used, so this is just a comment (in case it's important); note that on the containers itself, environment variables are set as a `[]string`. IIRC, one of the reasons was that order may be important (and Go maps randomise order when iterating over them).

For example; this is valid, and later values for `FOO` can override earlier values;

```
docker run -e FOO=bar -e FOO=baz -e BLA=BLA
```

If there's an intermediate map, we may loose the original.

Depending on "when" env-vars are expanded, and which values should be taken into account, `BLA=$FOO` may have to be substituted, and same for `-e BLA` (which takes the value of `$BLA` from the current environment (if present). Without looking deeply into the code, this may mean that we may have to distinguish `nil` (`-e FOO`) from `empty` values (`-e FOO=` / `-e FOO=""`).

cli/options.go
```diff
@@ -37,53 +37,84 @@ type ProjectOptions struct {
...
 
-	return ProjectOptions{
-		Name:        o.Name,
-		WorkingDir:  o.WorkingDir,
-		ConfigPaths: o.ConfigPaths,
-		Environment: getAsStringList(env),
+// WithName defines ProjectOptions' name
+func WithName(name string) ProjectOptionsFn {
+	return func(o *ProjectOptions) error {
+		o.Name = name
```
thaJeztah, line 63: Should validation take place at this point, or is that done later?

ndeloof: could be done here. So far we haven't clarified name constraints

cli/options.go
ndeloof: The code compose-go was inherited from also eventually uses a `map[string]string` after passing CLI options as a `[]string`: https://github.com/docker/cli/blob/master/cli/compose/types/types.go#L63

thaJeztah: Thx! So I guess if there _is_ a problem in that approach, it's not something _new_ 👍

PR #62, opened 2020-07-17
Merge V2 - V3 compose file formats (optional version field)

ndeloof:
This is the compose-go sibling PR for https://github.com/docker/compose/pull/7588 to apply https://github.com/compose-spec/compose-spec/issues/13

Note: this PR **don't** add support for compose file v1, something we could consider later


Papipo:
Shouldn't this allow me to use `extends` like I did in `docker-compose`? Because now I get this error in a service that `extends` (I am also using multiple `-f` flags):

`service "superset_deployment" has neither an image nor a build context specified: invalid compose project`

PR #68, opened 2020-08-20
Manage relative paths ( `./*` , `~/*` and `~user/*` )

ndeloof:
This adds support for config using `~/...` syntax for paths relative to user's HOME


cli/options.go
```diff
@@ -283,3 +295,15 @@ func getAsStringList(em map[string]string) []string {
 	}
 	return m
 }
+
+// resolvePath resolves "~" in paths
+func resolvePath(path string) (string, error) {
+	if strings.HasPrefix(path, "~") {
```
rumpl, line 301: Shouldn't this be `"~/"` since we are doing `path[2]` later

ndeloof: should be "~"+filepath.Separator then

rumpl: Yah, and `path[2]` should be... not sure what... I'm not fluent in windows paths but does `~\\test` mean something?

cli/options_test.go
```diff
@@ -84,3 +85,20 @@ func TestEnvMap(t *testing.T) {
...
+	assert.NilError(t, err)
+	assert.Equal(t, path, "")
+
+	path, err = resolvePath("/tmp/foo")
+	assert.NilError(t, err)
+	assert.Equal(t, path, "/tmp/foo")
+
+	path, err = resolvePath("~/foo")
+	assert.NilError(t, err)
+	assert.Equal(t, path, filepath.Join(home, "foo"))
```
justincormack, line 103: you should test plain "~" too.

cli/options.go
justincormack: should this work on Windows at all? This is not AFAIK a platform convention. Although WSL2 complicates things.

rumpl: It would seem `cd ~` works on PowerShell

ndeloof: Actually, I wonder we have to manage this _at all_. Isn't this a shell extension thing, that should take place BEFORE our code get ran ?

rumpl: It's a shell extension thing yes

ndeloof: answering myself: we need this to support use of `~/...` inside yaml as referenced files

loader/paths.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+)
+
+// ExpandPath resolves "~" and "." in paths to produce absolute paths
+func ExpandPath(workingDir string, path string) (string, error) {
```
rumpl, line 27: nit: why not name the function `ResolvePath` or simply `Resolve` since it's what this does

ndeloof: cause it's based on docker-compose's `expand_path`

loader/paths_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"gotest.tools/v3/assert"
+)
+
+func TestExpandPath(t *testing.T) {
+	workingDir := "/foo"
+	home, err := os.UserHomeDir()
+	assert.NilError(t, err)
+
+	path, err := ExpandPath(workingDir, "")
```
silvin-lubecki, line 33: nit: I guess you can use table driven test here

loader/paths_test.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, tt := range tests {
+		t.Run(tt.path, func(t *testing.T) {
+			got, err := ExpandPath(workingDir, tt.path)
+			assert.NilError(t, err)
+			assert.Equal(t, got, tt.expected)
+		})
+	}
+}
+
+func Abs(s string) string {
```
rumpl, line 90: nit: there is no need for this function to be exported

PR #72, opened 2020-09-01
Extends fixes

kklin:
These are various fixes that I've made on my fork of compose-go.

I'd loved to get them merged upstream. I'm happy to write tests, make changes, etc -- I assume I'll need to make some changes. I just made these changes in the moment based on the path of least resistance.

The last commit in particular feels ugly to me, especially since it changes the public API.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -441,10 +441,11 @@ func LoadServices(servicesDict map[string]interface{}, workingDir string, lookup
 				if err != nil {
 					return nil, err
 				}
-				source, err = ParseYAML(bytes)
+				parsedFile, err := ParseYAML(bytes)
 				if err != nil {
 					return nil, err
 				}
+				source = getSection(parsedFile, "services")[*service]
 			}
 			baseService, err := LoadService(name, source.(map[string]interface{}), workingDir, lookupEnv)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


loader/loader.go
```diff
@@ -414,51 +415,97 @@ func formatInvalidKeyError(keyPrefix string, key interface{}) error {
...
-				}
-				source, err = ParseYAML(bytes)
+		services = append(services, *serviceConfig)
+	}
+
+	return services, nil
+}
+
+func loadServiceWithExtends(name string, servicesDict map[string]interface{}, workingDir string, lookupEnv template.Mapping, opts *Options, depth int) (*types.ServiceConfig, error) {
+	if depth > 100 {
```
ndeloof, line 434: Would better track services we've already seen (like https://github.com/docker/compose/blob/master/compose/config/config.py#L635 does) to detect cycles. Could also track as a slice so we can report the cycle to end user for better diagnostic

kklin: Ah yeah I considered that, but wasn't sure how to uniquely identify the service. Do you think I should modify the functions to take the filename as an arg as well?

ndeloof: to mimic docker-compose this probably makes sense to refer to seen services as `filename:service` (see https://github.com/docker/compose/blob/master/compose/config/config.py#L585-L586). So passing filename to this func makes perfect sense to me

kklin: I'm not sure how to get the filename of config file that's initially passed to `loadSections`. It doesn't look like that info is available anywhere.

~~I could imagine using the empty string to represent that initial file's filename, but then we wouldn't be able to detect cyclic dependencies to that file.~~

Actually, I think that would work. We wouldn't detect the cycle on the first pass, but after loading the service the second time, it'd have the filename set. So I'll plan to write it like that.

kklin: @ndeloof I've added the improved cycle tracking

ndeloof (approved):

PR #74, opened 2020-09-07
support x-* extension in transformed types

ndeloof:
Some types get special handling during the loading process to manage short vs long syntax
when we process so, we need to invoke `groupXFieldsIntoExtensions` to populate the `Extensions` field from all `x-*` elements.


chris-crone (approved):
LGTM

PR #75, opened 2020-09-21
Add support for GPU specification as describe by compose spec

ndeloof:
As adopted by https://github.com/compose-spec/compose-spec/pull/100


chris-crone (approved):
LGTM but CI is sad

PR #79, opened 2020-10-29
Ignore double quotes on parsing types.Duration from JSON

ulyssessouza:
This is a fix to be integrated in https://github.com/docker/compose-cli/pull/845 for the `Healthcheck Interval`

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -52,6 +52,16 @@ func (d Duration) MarshalYAML() (interface{}, error) {
 	return d.String(), nil
 }
 
+func (d *Duration) UnmarshalJSON(b []byte) error {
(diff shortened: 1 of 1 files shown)


ndeloof:
AFAIK yaml support for this is implemented by [transformStringToDuration](https://github.com/compose-spec/compose-go/blob/ignore-double-quotes-duration/loader/loader.go#L1012-L1023), I wonder we could adopt the same approach having "marshaller-aware types" so that we don't need such custom transform functions. That's weird to me we introduce this Unmarshal method for JSON without having the same one for YAML (and maybe could share code).

gtardif (approved):

PR #92, opened 2020-12-14
Check service has either a Build section or Image set

ndeloof:
Detect invalid image definition for service (no build, no image) as we load compose file

see https://github.com/docker/compose-cli/issues/1055


gtardif (approved):

gtardif:
This will allow removing a similar check in compose-cli build.go

PR #93, opened 2021-01-05
Enable `all` as DeviceRequest.Count value and fix typo

aiordache:
Enable `all` as `Count` value and change `Diver`-> `Driver` in DeviceRequest struct.

Closes https://github.com/docker/dev-tooling-team/issues/264

Signed-off-by: aiordache <anca.iordache@docker.com>

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -377,8 +377,8 @@ type Resource struct {
 
 type DeviceRequest struct {
 	Capabilities []string `mapstructure:"capabilities" yaml:"capabilities,omitempty" json:"capabilities,omitempty"`
-	Diver        string   `mapstructure:"driver" yaml:"driver,omitempty" json:"driver,omitempty"`
-	Count        int      `mapstructure:"count" yaml:"count,omitempty" json:"count,omitempty"`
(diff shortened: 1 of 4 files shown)


types/types.go
```diff
@@ -377,8 +377,8 @@ type Resource struct {
 
 type DeviceRequest struct {
 	Capabilities []string `mapstructure:"capabilities" yaml:"capabilities,omitempty" json:"capabilities,omitempty"`
-	Diver        string   `mapstructure:"driver" yaml:"driver,omitempty" json:"driver,omitempty"`
-	Count        int      `mapstructure:"count" yaml:"count,omitempty" json:"count,omitempty"`
+	Driver       string   `mapstructure:"driver" yaml:"driver,omitempty" json:"driver,omitempty"`
+	Count        int64    `mapstructure:"count" yaml:"count,omitempty" json:"count,omitempty"`
```
ndeloof, line 381: While fixing this typo, please note `count` can also be set to `all` and should be of type string

ndeloof (approved):

PR #94, opened 2021-01-13
Fix network naming in normalization

gtardif:
If not specified in yaml, network names are normalized to `project_networkKey` , this is the final network name used to create network by docker-compose or compose-cli implementation. 
Previously it was normalized to just `networkKey` and later on we have no way to know if this was specified by user or not, when adding the `project_` prefix.

Same for volumes, config & secrets for consistency. 
This impact downstream projects : there is no additional name manipulation needed to add a prefix

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -229,60 +229,35 @@ func loadSections(filename string, config map[string]interface{}, configDetails 
 		Filename: filename,
 	}
 
-	var loaders = []struct {
-		key string
-		fnc func(config map[string]interface{}) error
-	}{
-		{
-			key: "services",
-			fnc: func(config map[string]interface{}) error {
-				cfg.Services, err = LoadServices(filename, config, configDetails.WorkingDir, configDetails.LookupEnv, opts)
-				return err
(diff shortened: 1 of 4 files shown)


loader/loader.go
```diff
@@ -229,60 +229,35 @@ func loadSections(filename string, config map[string]interface{}, configDetails
...
 	}
-	for _, loader := range loaders {
-		if err := loader.fnc(getSection(config, loader.key)); err != nil {
-			return nil, err
-		}
+
+	cfg.Networks, err = LoadNetworks(getSection(config, "networks"), configDetails.Version)
+	if err != nil {
+		return nil, err
+	}
```
gtardif, line 240: Unfortunately I still have the go error checking for each line here...

ndeloof: you can do 
```go
err := firstErr(LoadNetworks, LoadVolume, ...)
```
with 
```go
func firstErr(fns ... func(map[string]interface{}, types.ConfigDetails, *types.Config)) error
```

PR #102, opened 2021-02-03
Merge top-level extensions

ndeloof:
fixed `merge` to also handle top-level `x-*` extension fields


gtardif (approved):
except for red test(nil / empty extension), LGTM

PR #104, opened 2021-02-17
introduce Project.ResolveImages to use image digests

ndeloof:
Passing a resolver func, this allows a compose project to use digested image references


chris-crone (approved):
LGTM

PR #105, opened 2021-02-17
embed schema.json using go 1.16 support for static resources

ndeloof:
The `embed` package add support for embedding static resources without relying on code generation
https://golang.org/pkg/embed/


.github/workflows/ci.yml
```diff
@@ -16,13 +16,13 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.13.x, 1.14.x]
+        go-version: [1.16.0]
```
chris-crone, line 19: We should make this float with patch releases.

```suggestion
        go-version: [1.16.x]
```

schema/compose-spec.json
```diff
@@ -87,7 +87,7 @@
                 "dockerfile": {"type": "string"},
                 "args": {"$ref": "#/definitions/list_or_dict"},
                 "labels": {"$ref": "#/definitions/list_or_dict"},
-                "cache_from": {"$ref": "#/definitions/list_of_strings"},
+                "cache_from": {"type": "array", "items": {"type": "string"}},
```
chris-crone, line 90: Were the changes to the schema intentional?

ndeloof: I used the latest from compose-spec repo

chris-crone (approved):
LGTM with CI fix

PR #106, opened 2021-02-17
Workaround mergo merging panic for volumes

ulyssessouza:
This should fix https://github.com/docker/compose-cli/issues/1261 and https://github.com/docker/cli/issues/2916

The process should probably be applied to other maps, like `networks`, `secrets`, `configs` and `extensions`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.4.0
 	github.com/google/go-cmp v0.5.4
-	github.com/imdario/mergo v0.3.11
+	github.com/imdario/mergo v0.3.12-0.20210303221052-29fb3d3bdc55
 	github.com/joho/godotenv v1.3.0
 	github.com/mattn/go-shellwords v1.0.11
 	github.com/mitchellh/mapstructure v1.4.1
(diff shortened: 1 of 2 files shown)


ndeloof:
If we adopt the "pointer" approach, then we need to do the same on all `map[]foo` types

ulyssessouza:
@ndeloof You are right!
This is just a "Draft" PR to check if there is something missing before I do the rest of the types

ndeloof:
Reading https://github.com/imdario/mergo/issues/90#issuecomment-435606388 it seems this would not be the sivler bullet you expected: non-empty source would be fully replaced by override, not merged.

ndeloof:
I ran my own debug session investigating https://github.com/docker/compose-cli/issues/1261 ...
surprisingly, mapstruct is able to set a `map` field (see https://github.com/mitchellh/mapstructure/blob/master/mapstructure.go#L846) while mergo fails doing the same https://github.com/imdario/mergo/blob/cd55aeac72a19c6d5826061d5125d72b252a12eb/merge.go#L104
but that's all voodoo to me (ㆆ _ ㆆ)

I noticed (trying to get a minimal test set) that issue happens as compose override file defines `driver_opts` while the source has this map initial value set to nil
i.e adding an empty map in the yaml doc fix this issue:
```
volumes:
  data:
    driver_opts: {}
```
so a _possible_ fix, without digging into mergo internal reflect voodoo is to enforce all maps to be set.
mapstructure's `ZeroFields` could help us here, but unfortunately this only applies to keys in source, and other fields are skipped (https://github.com/mitchellh/mapstructure/blob/master/mapstructure.go#L1350-L1353)
Need more investigations :P

ndeloof (changes requested):
tested with 
```
volumes:
  data:
```
+ override:
```
volumes:
  data:
    driver: azure_file
    driver_opts:
      share_name: test-share
      storage_account_name: my-storage-account
```
as reported in initial issue

=> panic persists

by my own investigation, root cause is `DriverOpts` is a nil map and mergo fails to set a value.
Can't find a satisfying fix so far.

ulyssessouza:
@ndeloof You are right!
This is just a "Draft" PR to check if there is something missing before I do the rest of the types

ulyssessouza:
@ndeloof  Just waiting on merge of https://github.com/imdario/mergo/pull/177 to bump `mergo` to fix the `nil` issue

ndeloof:
@ulyssessouza https://github.com/imdario/mergo/pull/177 has been merged

ulyssessouza:
@ndeloof Yep! And I've just bumped mergo in this PR too

ndeloof:
ack, but with the fix you should not need to use `*types.VolumeConfig` anymore

ndeloof (approved):

PR #109, opened 2021-02-18
Validate network_mode for service and container

ulyssessouza:
Note that it also refactors `Project::GetServices` and `Project::GetService`

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -93,30 +93,56 @@ func (p Project) ConfigNames() []string {
 	return names
 }
 
+func (p Project) GetByContainerName(names ...string) (Services, error) {
+	if len(names) == 0 {
+		return p.Services, nil
+	}
+	services := Services{}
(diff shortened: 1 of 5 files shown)


types/project.go
```diff
@@ -93,30 +93,56 @@ func (p Project) ConfigNames() []string {
...
+			}
+		}
+		return nil, fmt.Errorf("service with container_name %q could not be found", name)
+	}
+	return services, nil
+}
+
 // GetServices retrieve services by names, or return all services if no name specified
-func (p Project) GetServices(names []string) (Services, error) {
+func (p Project) GetServices(names ...string) (Services, error) {
```
ulyssessouza, line 115: Note this refactor to vargs. `GetService` was kept as a comfort function to avoid the collection.

ndeloof (approved):

PR #112, opened 2021-03-02
introduce func to build Mapping from a set of KEY=VALUE strings

ndeloof:


gtardif (commented):
Could use unit tests on this kind of conversion methods 👼 
Can't remember why specifically we needed both types of mapping (string and *string) ; but anyway, no pb.

gtardif (approved):

PR #120, opened 2021-04-12
add blkio_config struct to match compose-spec schema

ndeloof:
struct is defined in schema but was mapped to a `string` by legacy docker/cli/compose implementation


aiordache (approved):

chris-crone (approved):
LGTM

PR #125, opened 2021-04-20
get COMPOSE_* from env sources as set by options

ndeloof:
fixes https://github.com/docker/compose-cli/issues/1570

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -223,7 +223,6 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 
 // getConfigPathsFromOptions retrieves the config files for project based on project options
 func getConfigPathsFromOptions(options *ProjectOptions) ([]string, error) {
-	paths := []string{}
 	pwd := options.WorkingDir
 	if pwd == "" {
 		wd, err := os.Getwd()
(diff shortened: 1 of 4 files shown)


gtardif (changes requested):
LGTM but can you add a test next to this ?

aiordache (approved):

cli/options.go
```diff
@@ -290,6 +269,24 @@ func getConfigPathsFromOptions(options *ProjectOptions) ([]string, error) {
...
+			paths = append(paths, f)
+			continue
+		}
+		if !filepath.IsAbs(f) {
+			f = filepath.Join(pwd, f)
+		}
+		if _, err := os.Stat(f); err != nil {
+			return paths, err
+		}
+		paths = append(paths, f)
```
gtardif, line 285: Why not transforming this into an absolute path anymore here ?

ndeloof: because we just did on line 279
I'm assuming here we pass an absolute `pwd`, which should be either os.Gwd or project.WorkingDir

gtardif: ok, as the PR for absolute paths on working dir has been merged

gtardif (approved):

PR #126, opened 2021-04-20
resolve build args and context path during `normalize` phase

ndeloof:
Get compose model loading to resolve build context and Dockerfile paths into absolute paths, and resolve build args from configured environment variables (depends on configured sources: os.Env, .env, --env-file)
this will make Compose implementation code simpler as they don't have to handle this

see https://github.com/docker/compose-cli/issues/1540


chris-crone (approved):
LGTM

PR #127, opened 2021-04-22
shm_size can be set as pain int or unitByte notation

ndeloof:
shm_size can be set both as plain int or a string with byte notation

fix https://github.com/docker/compose-cli/issues/1565


gtardif (approved):

gtardif:
why this CI error that's not related to the current change ??

aiordache (approved):

PR #129, opened 2021-04-23
resolve paths relative to '~', reduce code duplication with absPath

ndeloof:
fix https://github.com/docker/compose-cli/issues/1584


gtardif (approved):
LGTM. Probably worth trying to integrate this in a compose-cli PR and run e2e tests to validate it does not bring unwanted side effects...

ndeloof:
yes indeed, so https://github.com/docker/compose-cli/pull/1585

PR #131, opened 2021-04-26
introduce WithDefaultConfigPath and support for overrides

ndeloof:
compose-go used to search for compose.yaml in parent folders, while this is not defined by the spec. Also miss support for override files : https://github.com/docker/compose-cli/issues/1546

With proposed changes, looking for compose.yaml and the sibling override file is an opt-in feature

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -70,8 +70,49 @@ func WithName(name string) ProjectOptionsFn {
 // WithWorkingDirectory defines ProjectOptions' working directory
 func WithWorkingDirectory(wd string) ProjectOptionsFn {
 	return func(o *ProjectOptions) error {
-		o.WorkingDir = wd
+		if wd == "" {
+			return nil
+		}
+		abs, err := filepath.Abs(wd)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


cli/options.go
```diff
@@ -70,11 +70,52 @@ func WithName(name string) ProjectOptionsFn {
 // WithWorkingDirectory defines ProjectOptions' working directory
 func WithWorkingDirectory(wd string) ProjectOptionsFn {
 	return func(o *ProjectOptions) error {
+		if !filepath.IsAbs(wd) {
+			abs, err := filepath.Abs(wd)
```
mat007, line 74: nit: any reason to test beforehand? Isn’t `filepath.Abs` a no-op if it’s already absolute?

gtardif: Actually it's not totally, on windows for example "/xx/yyy" would return true to `filepath.IsAbs()` but `filepath.Abs()` would update the path as a windows path, with drive etc. (always something fishy with windows paths 😁  )
That being said, maybe it's not an issue to still resolve `filepath.Abs()` regardless.

gtardif (approved):

PR #132, opened 2021-04-27
introduce CreateHostPath

ndeloof:
introduce Bind option `CreateHostPath` and set automatically when service volume has been parsed from short syntax

### DIFF
--- a/loader/volume.go
+++ b/loader/volume.go
@@ -119,18 +119,26 @@ func isBindOption(option string) bool {
 }
 
 func populateType(volume *types.ServiceVolumeConfig) {
-	switch {
-	// Anonymous volume
-	case volume.Source == "":
-		volume.Type = string(types.VolumeTypeVolume)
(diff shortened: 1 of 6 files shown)


loader/volume.go
```diff
@@ -119,18 +119,25 @@ func isBindOption(option string) bool {
...
-	case isFilePath(volume.Source):
-		volume.Type = string(types.VolumeTypeBind)
-	default:
-		volume.Type = string(types.VolumeTypeVolume)
+	if isFilePath(volume.Source) {
+		volume.Type = types.VolumeTypeBind
+		if volume.Bind == nil {
+			volume.Bind = &types.ServiceVolumeBind{}
+		}
+		// Using short notation on bind involves bind will create missing host path
```
gtardif, line 127: Also comment why we do this, for compatibility with docker-compose

gtardif (approved):

PR #133, opened 2021-04-29
whenever backed by a ShellCommand, service.command must NOT merge WithAppendSlice

ndeloof:
service.Command is implemented as a `type ShellCommand []sring` and so mergo applies `WithAppendSlice` policy, which isn't relevant in this case.

close https://github.com/docker/compose-cli/issues/1601


loader/merge.go
```diff
@@ -247,6 +250,14 @@ func mergeServiceNetworkConfig(dst, src reflect.Value) error {
 	return nil
 }
 
+//nolint: unparam
+func mergeServiceCommand(dst, src reflect.Value) error {
```
gtardif, line 254: Some comment on this code would be useful to understand why it's there (some callback executed by reflection I guess).
Does it need to be added [here](https://github.com/compose-spec/compose-go/blob/master/loader/merge.go#L40) ? otherwise, how does this work ?

ndeloof: obsolete sorry, was my first attempt trying to fix this

gtardif (approved):

PR #142, opened 2021-06-04
run interpolation on raw yaml before parsing into map[string]interface{}

ndeloof:
This move the interpolation process to take place _before_ parsing the yaml tree into a
`map[string]interface{}`, so that a variable which resolves to a numeric value will corectly be parsed as an `int`, not `string`.

tested with https://github.com/docker/compose-cli/pull/1738

see https://github.com/docker/compose-cli/issues/1725 for context

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -140,15 +140,16 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 	}
 
 	configs := []*types.Config{}
-	for _, file := range configDetails.ConfigFiles {
+	for i, file := range configDetails.ConfigFiles {
 		configDict := file.Config
-
-		if !opts.SkipInterpolation {
-			var err error
-			configDict, err = interpolateConfig(configDict, *opts.Interpolate)
+		if configDict == nil {
+			dict, err := parseConfig(file.Content, opts)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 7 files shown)


loader/loader.go
```diff
@@ -442,6 +455,15 @@ func loadServiceWithExtends(filename, name string, servicesDict map[string]inter
 			baseFilePath := absPath(workingDir, *file)
 
 			bytes, err := ioutil.ReadFile(baseFilePath)
+
```
mat007, line 458: Should we still check `err` here maybe?

ulyssessouza (approved):
LGTM

slavdok:
What and where am I supposed to update to fix this issue? This was marked fixed in 2021. I've installed Docker and AWS CLI from scratch in 2022. I am having this exact same issue as described.

Server: Docker Desktop 4.12.0 (85629)
Engine:
Version: 20.10.17
Docker Compose version v2.10.2

milas:
@slavdok Please [open a new issue in `docker/compose`](https://github.com/docker/compose/issues/new?assignees=&labels=status%2F0-triage%2Ckind%2Fbug&template=bug_report.yml&title=%5BBUG%5D+%3Ctitle%3E) with the bug report template including a simplified repro (`compose.yaml` + commands to trigger).

PR #145, opened 2021-06-10
compose file path set from flags are relative to working directory

ndeloof:
when passed as command line parameter, `./xxx`  path should be considered relative to current working directory, whenever a project directory is set (which defines relative paths used within the compose file)

close https://github.com/docker/compose-cli/issues/1771


ulyssessouza (approved):
LGTM

PR #146, opened 2021-06-12
use GetWorkingDir instead of WorkingDir

satotake:
Execution of `docker compose up` in sub dir throws `can't find a suitable configuration file in this directory or any parent`,
even if `docker-compose.yml` is placed in parent dir properly.

If you do not pass cli the option and `o.WorkingDir` is empty, `filepath.Dir(pwd)` returns "." as parent dir.
Fix this by using `o.GetWorkingDir`


ndeloof:
Thanks for looking into this!

ndeloof (approved):

PR #147, opened 2021-06-14
relative paths for CLI options must be resolved based on pwd

ndeloof:
Resolve relative paths set by CLI options based on current working directory, not _project directory_ that should be used only for relative paths within the compose file.

close https://github.com/docker/compose-cli/issues/1780


ulyssessouza (approved):
LGTM

PR #149, opened 2021-06-23
drop unnecessary resources from project model

ndeloof:
This allows to remove any resource (network, volume, ...) that is not required by the active set of services, including the implicit "default" network when services are explicitly bound to declared networks

close https://github.com/docker/compose-cli/issues/1756

Proof of usability: https://github.com/ndeloof/compose-cli/runs/2893113262


ulyssessouza (approved):
LGTM

PR #151, opened 2021-06-28
fix `ports: []` fail to validate schema

ndeloof:
`[]` must be converted into an empty list, not `nil` otherwise schema validation fail


mat007 (approved):
Makes sense!

ulyssessouza (approved):
LGTM

PR #152, opened 2021-06-30
assume remote context if local path doesn't exists

ndeloof:
fix https://github.com/docker/compose-cli/issues/1854

I wish we could introduce some `IsRemoteContext` func based on a well defined set of requirements for remote context URLs, but according to https://github.com/moby/moby/issues/33686#issuecomment-308748173 there's a large set of de facto hack-ish syntax being supported by `docker build` we'd like to embrace.

So better assume the compose implementation/container runtime will be able to validate path/URL and report adequate error when relevant.

### DIFF
--- a/loader/normalize.go
+++ b/loader/normalize.go
@@ -18,7 +18,7 @@ package loader
 
 import (
 	"fmt"
-	"net/url"
+	"os"
 	"path/filepath"
 
 	"github.com/compose-spec/compose-go/errdefs"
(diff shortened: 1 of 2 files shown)


loader/normalize.go
```diff
@@ -70,10 +70,13 @@ func normalize(project *types.Project) error {
 			if s.Build.Dockerfile == "" {
 				s.Build.Dockerfile = "Dockerfile"
 			}
-			url, err := url.ParseRequestURI(s.Build.Context)
-			if err != nil || url.Scheme == "" {
-				s.Build.Context = absPath(project.WorkingDir, s.Build.Context)
-				s.Build.Dockerfile = absPath(s.Build.Context, s.Build.Dockerfile)
+			localContext := absPath(project.WorkingDir, s.Build.Context)
+			if _, err := os.Stat(localContext); err == nil {
```
gtardif, line 74: maybe stat Dockerfile rather than just the context folder ?

gtardif (approved):
Ah OK, forget stating the Dockerfile, we want to check the context is a local one and not a git remote context. LGTM

PR #154, opened 2021-07-02
force project name to be lowercase

ndeloof:
This force project name to be lowercase.
Required as service image name is inferred by project name, and must be lowercase
see https://github.com/docker/compose/blob/4a51af09d6cdb9407a6717334333900327bc9302/compose/cli/command.py#L197-L198

close https://github.com/docker/compose-cli/issues/1871


ndeloof:
my bad, project name set by compose file parent dir is already converted to lower case

ulyssessouza (approved):
LGTM

PR #155, opened 2021-07-02
warn on unset variable

ndeloof:
Warn user that compose.yaml includes variable which were not resolved, and an empty string will be used.

close https://github.com/docker/compose-cli/issues/1872


caugner:
@ndeloof Unfortunately this also warns for variables that are explicitly used with a default value, e.g. `${UID:-1000}`

PR #158, opened 2021-07-22
drop all comments from yaml file before applying string substitutions

ndeloof:
fix https://github.com/docker/compose-cli/issues/1896


chris-crone (approved):
LGTM

PR #159, opened 2021-07-22
warn for unset variable only during interpolation

ndeloof:
close docker/compose-cli#1918

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+
+	"github.com/sirupsen/logrus"
 )
 
 var delimiter = "\\$"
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
Overall looks OK to me code-wise, but is there a meaningful way to test this change?

template/template.go
```diff
@@ -97,7 +99,10 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
 			return value
 		}
 
-		value, _ := mapping(substitution)
+		value, ok := mapping(substitution)
+		if !ok {
+			logrus.Warnf("The %s variable is not set. Defaulting to a blank string.", substitution)
```
ulyssessouza, line 104:
```suggestion
			logrus.Warnf("The %q variable is not set. Defaulting to a blank string.", substitution)
```

ndeloof:
just noticed I already pushed this one as https://github.com/compose-spec/compose-go/pull/157

ulyssessouza (approved):
LGTM

PR #163, opened 2021-08-05
[Bug] Reproductive case

lewiscowleschipuk:
This replicates a bug found by Colleague @MrHuli when using `extends` in a single file; the commands seem to be concatenated, even if this is not the behaviour of docker-compose v1 branch.

Can someone explain to me if this is expected and to the spec, or needs fixing? Test is here. Just figured the reproduction would be more use than an issue (in case it is not an issue).

### Update
I found that merge is not the component responsible (I have a test for loader too locally); I Think given the past issue, it makes sense to me to keep a check in merge too.

I Can contribute the loader case if needed, ~but I think this may actually fix merge.go to not need~ (confirmed merge.go needs its fix too). It simply checks that two commands merge into one if the extending service also defines a command.

This does seem a shame, but I can see a lot of breaking backwards compatibility if people play with `extends`.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -516,10 +516,14 @@ func loadServiceWithExtends(filename, name string, servicesDict map[string]inter
 			}
 		}
 
+		commandBackup := serviceConfig.Command
 		if err := mergo.Merge(baseService, serviceConfig, mergo.WithAppendSlice, mergo.WithOverride, mergo.WithTransformers(serviceSpecials)); err != nil {
 			return nil, errors.Wrapf(err, "cannot merge service %s", name)
 		}
 		serviceConfig = baseService
+		if len(commandBackup) > 0 {
+			serviceConfig.Command = commandBackup
+		}
 	}
 
 	return serviceConfig, nil
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -290,6 +290,25 @@ services:
 	assert.Check(t, service.Command[0] == "echo")
 }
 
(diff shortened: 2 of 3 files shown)


ndeloof:
AFAIK this has allready been fixed by https://github.com/compose-spec/compose-go/pull/133

Lewiscowles1986:
🙄 please remove non test cases and watch it fail... The broken test was literally thing 1 committed.

There is an odd behaviour when a single yaml file is used as shown in the tests.

ndeloof:
sure, I'm just saying  #133 _should_ have fixed this.

Lewiscowles1986:
I thought the same thing. Notice this patches loader. I also had hoped that would patch this, but it seems that both use the same merging library without a rule to skip `command`. It's a sucky hack. What is weird is that it works on my local, but not here 😬

ndeloof:
Also notice we have to do the same for `Entrypoint`

Lewiscowles1986:
So to add to the weird, this is working in osx, but I believe may need the order switched in linux (only reason I can fathom it would fail is if `baseService` and `serviceConfig` were being muddled. 😞 . It still works on my local on two separate machines. I won't be able to test windows or linux until later.

Lewiscowles1986:
@ndeloof what does this mean?
> Also notice we have to do the same for Entrypoint

ndeloof:
Created https://github.com/compose-spec/compose-go/pull/164, which included proposed reproduction test case

PR #167, opened 2021-08-17
Remove only full line comments to avoid variable interpolation on this ones

ulyssessouza:
This is due to the Marshal and Unmarshal methods
also remove double quotes from strings.
This breaks empty strings turning it into `nil` values.

PTAL @lorenrh

Signed-off-by: Ulysses Souza <ulyssessouza@gmail.com>

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


lorenrh (approved):
Thanks!

StefanScherer (approved):
LGTM

rumpl (approved):

ndeloof:
LGTM but then we need another way to remove comments :'(

ndeloof:
I guess unmarshall+marshall way to drop comments also caused https://github.com/docker/compose-cli/issues/2036

ndeloof:
I suggest `removeYamlComments` is kept but implementation just drop lines starting by `#`. This won't remove all comments, but at least the most obvious ones where users might have documented usages with some command line examples

ndeloof (approved):

Issue #168, opened 2021-08-23
Export loadOptions?


michaelperel:
When using [`cli.ProjectFromOptions`](https://github.com/compose-spec/compose-go/blob/b2976a566ee80ac98a8b1f1fe0705233f6314eb5/cli/options.go#L267), the functional options pattern allows you to specify behavior for every exported field of [`cli.ProjectOptions`](https://github.com/compose-spec/compose-go/blob/b2976a566ee80ac98a8b1f1fe0705233f6314eb5/cli/options.go#L36).

However, there is one field, `loadOptions`, which is not exported. This is problematic for me because I would like to use `cli.ProjectFromOptions`, but I want to specify that the `loader` should skip validation.

Would a PR where either
(1) `loadOptions` is exported so it can be controlled via functional options like the rest of the fields

or

(2) the loader is specified to `cli.ProjectFromOptions` via dependency injection

be accepted?

Otherwise, I would have to use `loader.Load` and repeat logic that already exists via `cli.ProjectFromOptions`.

Thanks!

michaelperel:
Closed by #170

PR #170, opened 2021-08-24
load options when creating a new project

michaelperel:
This a POC demonstrating how we could expose load options, as in #168 

Here, I have provided functionality for one option, `SkipValidation`. If this approach is accepted, I could write the code for all the other options.

Here is an example of how it is used in [docker-lock](https://github.com/safe-waters/docker-lock/blob/1138e6e5db6d53cb66d8dee6c02f8242de636c9c/pkg/generate/parse/compose.go#L105).

I would love feedback / thoughts.

Thanks!

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -157,9 +157,7 @@ func WithDiscardEnvFile(o *ProjectOptions) error {
 // WithLoadOptions provides a hook to control how compose files are loaded
 func WithLoadOptions(loadOptions ...func(*loader.Options)) ProjectOptionsFn {
 	return func(o *ProjectOptions) error {
-		for _, f := range loadOptions {
-			o.loadOptions = append(o.loadOptions, f)
-		}
+		o.loadOptions = append(o.loadOptions, loadOptions...)
 		return nil
 	}
 }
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

michaelperel:
@ndeloof I have updated it so that it should pass the failed linter & sign tests (but I need approval for them to run). I also removed the “WithSkipValidation” function in loader. Rather than specifying all the functions there, by just providing the hook, anyone can specify any function easily.

Here is [an example](https://github.com/safe-waters/docker-lock/blob/84f70e6e3a67f5cf7aa10741661017c9e29e38bb/pkg/generate/parse/compose.go#L111) of how it is used now.

If you would prefer the other way (making functions for clients to use in loader) please let me know. If so, do you want all the options specified via functions, or can I just leave the one function `WithSkipValidation` as originally in the PR?

ndeloof:
WithSkipValidation sounds good, let's make this simple for code consumers

michaelperel:
Great, I have added it back. Please let me know if there is anything else I need to do to get it merged :)

michaelperel:
@ndeloof please let me know if there is anything that I can do to get this merged. Thanks!

PR #173, opened 2021-08-31
Fix project name regexp

ulyssessouza:
This fixes the regexp on the case of a single character project name. 
Resolves https://github.com/docker/compose-cli/issues/2052

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -314,7 +314,7 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		} else if nameFromEnv, ok := options.Environment[ComposeProjectName]; ok && nameFromEnv != "" {
 			opts.Name = nameFromEnv
 		} else {
-			opts.Name = regexp.MustCompile(`(?m)[a-z]+[-_a-z0-9]*[a-z0-9]+`).FindString(strings.ToLower(filepath.Base(absWorkingDir)))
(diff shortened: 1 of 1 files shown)


cli/options.go
```diff
@@ -314,7 +314,7 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		} else if nameFromEnv, ok := options.Environment[ComposeProjectName]; ok && nameFromEnv != "" {
 			opts.Name = nameFromEnv
 		} else {
-			opts.Name = regexp.MustCompile(`(?m)[a-z]+[-_a-z0-9]*[a-z0-9]+`).FindString(strings.ToLower(filepath.Base(absWorkingDir)))
+			opts.Name = regexp.MustCompile(`(?m)[a-z]+[-_a-z0-9]*[a-z0-9]*`).FindString(strings.ToLower(filepath.Base(absWorkingDir)))
```
ndeloof, line 317: `[-_a-z0-9]*[a-z0-9]*̀  is redundant, could be just `[-_a-z0-9]*`

ndeloof (changes requested):

cli/options.go
ulyssessouza: Thats right! Thx!

ndeloof (approved):

PR #175, opened 2021-08-31
add support for `*` as a hack to enable all profiles

ndeloof:
Allows to enable all profiles declared in a compose file using special profile `*`

To be used as `docker compose --profile '*' ... `

#fixes https://github.com/docker/compose-cli/issues/1964


ulyssessouza (approved):
LGTM

PR #176, opened 2021-09-01
don't resolve `FOO=`, value is empty string

ndeloof:
variable set as `FOO=` should not be resolved based on project environment, but set to empty string

This is a fix https://github.com/docker/compose-cli/issues/1989


ulyssessouza (approved):
LGTM

PR #177, opened 2021-09-01
include IPC,PID,Volumes_From in dependencies detection

ndeloof:
GetDependencies doesn't consider depencies to services by `ipc: service:xx`, `pid: service:xx`, or `volumes_from: [xx]`, 

required (but also need a fix in compose.v2) for https://github.com/docker/compose-cli/issues/2035


ulyssessouza (approved):
LGTM

PR #178, opened 2021-09-03
introduce WithResolvedPaths to let client code opt for relative or abs

ndeloof:
Add a load option to let client code decide if paths should be resolved (absolute) or kept relative.
This addresses https://github.com/docker/compose/issues/8503


lorenrh (approved):
LGTM!

PR #179, opened 2021-09-09
fix merging ports

ndeloof:
fix port merge logic: align with docker-compose to use composite key `(target,publish,ip,protocol)` 
see https://github.com/docker/compose/blob/c81046aac0ce1360f1bab64368f9a318c593d82a/compose/config/types.py#L419

close https://github.com/docker/compose/issues/8557
close https://github.com/docker/compose/issues/8586


ulyssessouza (approved):
LGTM

PR #182, opened 2021-09-13
allow override by empty command

ndeloof:
`command` and `entrypoint` can be defined as blank string, as "unset shell command".
This still should apply as a valid override while merging compose files

close https://github.com/docker/compose/issues/8536


mat007 (approved):
Makes sense!

PR #183, opened 2021-09-14
transformSize accept UnitBytes created by interpolation

ndeloof:
interpolation produces the expected `UnitBytes` type for shm_size and comparable fields
this PR updates `transformSize` so it is a no-op when type is already correct

close https://github.com/docker/compose/issues/8613

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -1005,10 +1005,13 @@ var transformSize TransformerFunc = func(value interface{}) (interface{}, error)
 	switch value := value.(type) {
 	case int:
 		return int64(value), nil
+	case int64, types.UnitBytes:
+		return value, nil
 	case string:
 		return units.RAMInBytes(value)
+	default:
+		return value, errors.Errorf("invalid type for size %T", value)
 	}
-	panic(errors.Errorf("invalid type for size %T", value))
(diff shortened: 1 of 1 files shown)


loader/loader.go
```diff
@@ -1005,6 +1005,10 @@ var transformSize TransformerFunc = func(value interface{}) (interface{}, error)
 	switch value := value.(type) {
 	case int:
 		return int64(value), nil
+	case int64:
+		return value, nil
+	case types.UnitBytes:
+		return value, nil
```
mat007, line 1011: nit: It think this may work
```suggestion
	case int64, types.UnitBytes:
		return value, nil
```
?

mat007 (commented):
Should we add a ` default: `  and return an error to avoid possible panics in the future?

ndeloof:
That's indeed weird this TransformerFunc uses `panic` and doesn't just return an error, fixed

mat007 (approved):
Looks good!

toniopelo:
There might still be a problem @ndeloof, see https://github.com/docker/compose/issues/8613#issuecomment-1005850385. Might be a regression ?

PR #184, opened 2021-09-15
normalize scale -> deploy.replicas, defaults to 1

ndeloof:
- set Scale=1 by default before loading the service config from yaml
- normalize `scale` into `deploy.replicas` to conform to the compose specification
- serialize deploy.replicas even if set to `0` (to fix https://github.com/docker/compose/issues/8571)

close https://github.com/docker/compose/issues/8571


types/types.go
```diff
@@ -441,7 +441,7 @@ type LoggingConfig struct {
 // DeployConfig the deployment configuration for a service
 type DeployConfig struct {
 	Mode           string         `yaml:",omitempty" json:"mode,omitempty"`
-	Replicas       *uint64        `yaml:",omitempty" json:"replicas,omitempty"`
+	Replicas       *uint64        `yaml:"" json:"replicas"`
```
mat007, line 444: Is this change intentional? Other pointer fields in this struct do have `omitempty`.
I’m not sure what gets serialized in `Replicas` is `nil`.
Or maybe we’re only loading, so it doesn’t really matter?

loader/normalize.go
```diff
@@ -84,17 +84,22 @@ func normalize(project *types.Project, resolvePaths bool) error {
...
 		}
 
-		err = relocateDockerfile(s)
+		err = relocateDockerfile(&s)
+		if err != nil {
+			return err
+		}
+
+		err = relocateScale(&s)
 		if err != nil {
```
mat007, line 103: nit:
```suggestion
		if err := relocateScale(&s); err != nil {
```

types/types.go
ndeloof: yes it is, so that we get value serialized by `compose config` when set to `0`.
a `nil` value is never serialized.

loader/normalize.go
ndeloof: I hate this syntax, breaks readability ont he test condition imho

types/types.go
mat007: But `omitempty` here is for the pointer, if it’s non-`nil` and points to a `0` value, it will be serialized even if the field has `omitempty`.
Even if it has the same behaviour with or without `omitempty`, I think if only for consistency sake with the rest of the pointers, it should probably have an `omitempty`?

loader/normalize.go
mat007: However it saves on the few occasions when doing something like
```
err := something()
// forgetting to check err
if err := somethingElse(); err != nil {
}
```
This will make the compiler complain of an unused variable, whereas this doesn’t
```
err := something()
// forgetting to check err
err = somethingElse()
if err != nil 
{
}
```

But that’s fine, not a big deal.

types/types.go
ndeloof: `omitempty` does not apply to the pointer but to the actual value. nil pointer never get serialized.
Here we WANT `0` to be serialized (default value we would like to ignore is `1`)

ulyssessouza (approved):
LGTM

types/types.go
mat007: It’s the zero value of the field that matters (here it’s a pointer, so the zero value is `nil`): https://pkg.go.dev/gopkg.in/yaml.v2#Marshal (it’s v2 which I think compose-go is using, but it’s the same in v3):

> omitempty:    Only include the field if it's not set to the zero
             value for the type or to empty slices or maps.

`nil` pointers without `omitempty` yields a `null` value in the output: https://play.golang.org/p/Hk2qfCq9ccT
With `omitempty` the `nil` pointer gets omitted: https://play.golang.org/p/EVjf0jL2uN4
but not if the pointer is non-nil, even if it points to a 0 value: https://play.golang.org/p/7dype4Fh4cP

ndeloof: indeed. I was confused as `deploy` is istelf a pointer so was omitted.

ndeloof: so `replicas` is always serialized when non nil, and we only set value from `scale` by normalization if value != 1

Issue #118, opened 2021-04-08
Request: tag a release


AkihiroSuda:
For ease of dependency management, could you consider creating a release with git tag?

AkihiroSuda:
Linking related discussion : https://github.com/compose-spec/compose-spec/issues/104#issuecomment-816524801

alexellis:
We could really use a release here too @ndeloof for the usage in [faasd](https://github.com/openfaas/faasd)

Recent breaking changes in master the Go library have broken our usage, so we're having to pin to an old commit. It's harder to understand what changes are being made without any versions being cut, so what do you think about starting to create tags?

ndeloof:
I'd like we adopt something extra-simple to define tagged releases. https://github.com/compose-spec/compose-spec/issues/171 is still way too complex imho as it still require maintainers to agree on good time for a release...
I'd suggest we adopt a "Jenkins-style" weekly release cycle, i.e. have a cron job to tag and release `v1.xxx` wdyt?

> it's harder to understand what changes are being made without any versions being cut

AkihiroSuda:
https://github.com/compose-spec/compose-go/releases/tag/v1.0.1 🎉 
Thanks!

PR #185, opened 2021-09-15
weekly release workflow

ndeloof:
Automated weekly release workflow every Wednesday as `v1.{{week count}}.0` (jenkins-style)

example run on https://github.com/ndeloof/compose-go/releases/tag/v1.0.0
close https://github.com/compose-spec/compose-go/issues/118

(I would update workflow before merge so we get an initial release)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,21 @@
+name: Weekly Release
+
+on:
+  schedule:
+    - cron:  '12 0 * * WED'
+
+        
+jobs:
+  release:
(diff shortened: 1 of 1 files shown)


mat007 (commented):
Is there any way we can do a release only if there has been changes since the last one?
Maybe comparing the hash of the `master` branch head to the hash of the last release, or something like that?

ndeloof:
@mat007 we could do some `git log` magic for this purpose.
But as long as the version is based on counting weeks since xx, this will just skip some version number

PR #187, opened 2021-09-16
unbraced variable must not apply advanced substitutions

ndeloof:
close https://github.com/docker/compose/issues/8601

aligns with https://github.com/docker/compose/blob/2011bc3cea1c78eccd0fa75d01a6b638d6bfb7c2/compose/config/interpolation.py#L96-L98

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -25,11 +25,12 @@ import (
 )
 
 var delimiter = "\\$"
-var substitution = "[_a-z][_a-z0-9]*(?::?[-?][^}]*)?"
+var substitutionNamed = "[_a-z][_a-z0-9]*"
+var substitutionBraced = "[_a-z][_a-z0-9]*(?::?[-?][^}]*)?"
 
 var patternString = fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


template/template.go
```diff
@@ -25,11 +25,12 @@ import (
 )
 
 var delimiter = "\\$"
-var substitution = "[_a-z][_a-z0-9]*(?::?[-?][^}]*)?"
+var id = "[_a-z][_a-z0-9]*"
+var bid = "[_a-z][_a-z0-9]*(?::?[-?][^}]*)?"
```
mat007, line 29: What does `bid` mean?

mat007 (approved):
Makes sense!

template/template.go
ulyssessouza: Maybe naming it `substitutionNamed` and `substitutionBraced`

ndeloof: "braced ID" I guess?
My intent was to align with https://github.com/docker/compose/blob/2011bc3cea1c78eccd0fa75d01a6b638d6bfb7c2/compose/config/interpolation.py#L96-L98

ulyssessouza (approved):
nit: regexp naming.

PR #192, opened 2021-09-29
allow tmpfs.size to be set in unit bytes

ndeloof:
see https://github.com/compose-spec/compose-spec/pull/205

Close https://github.com/docker/compose/issues/8696

### DIFF
--- a/loader/full-struct_test.go
+++ b/loader/full-struct_test.go
@@ -403,7 +403,7 @@ func services(workingDir, homeDir string) []types.ServiceConfig {
 				{Source: "datavolume", Target: "/var/lib/mysql", Type: "volume", Volume: &types.ServiceVolumeVolume{}},
 				{Source: filepath.Join(workingDir, "opt"), Target: "/opt", Consistency: "cached", Type: "bind"},
 				{Target: "/opt", Type: "tmpfs", Tmpfs: &types.ServiceVolumeTmpfs{
-					Size: int64(10000),
+					Size: types.UnitBytes(10000),
 				}},
 			},
 			WorkingDir: "/code",
@@ -866,7 +866,7 @@ func fullExampleYAML(workingDir, homeDir string) string {
     - type: tmpfs
       target: /opt
       tmpfs:
-        size: 10000
+        size: "10000"
     working_dir: /code
     x-bar: baz
     x-foo: bar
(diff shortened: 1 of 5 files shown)


schema/compose-spec.json
```diff
@@ -426,8 +427,10 @@
                     "type": "object",
                     "properties": {
                       "size": {
-                        "type": "integer",
-                        "minimum": 0
+                        "oneOf": [
+                          {"type": "integer","minimum": 0},
```
mat007, line 431: nit:
```suggestion
                          {"type": "integer", "minimum": 0},
```

schema/compose-spec.json
```diff
@@ -599,12 +602,12 @@
...
-          "device_ids": {"$ref": "#/definitions/list_of_strings"},
-          "driver":{"type": "string"},
-          "options":{"$ref": "#/definitions/list_or_dict"}
-        },
+            "capabilities": {"$ref": "#/definitions/list_of_strings"},
+            "count": {"type": ["string", "integer"]},
+            "device_ids": {"$ref": "#/definitions/list_of_strings"},
+            "driver":{"type": "string"},
+            "options":{"$ref": "#/definitions/list_or_dict"}
+          },
```
mat007, line 610: Indentation seems to be 2 characters in the rest of the file, was this change intended?

loader/full-struct_test.go
```diff
@@ -866,7 +866,7 @@ func fullExampleYAML(workingDir, homeDir string) string {
     - type: tmpfs
       target: /opt
       tmpfs:
-        size: 10000
+        size: "10000"
```
mat007, line 869: Is this change intentional?

loader/loader_test.go
```diff
@@ -936,15 +936,15 @@ services:
       - type: tmpfs
         target: /app
         tmpfs:
-          size: 10000
+          size: 1kb
```
mat007, line 939: Do we want a test with both allowed formats? (or two tests)

loader/full-struct_test.go
ndeloof: yes, using UnitBytes makes the marshaller always prefer string

schema/compose-spec.json
ndeloof: I copied the exact content from compose-spec - we haven't yet found a way to directly refer to the reference repository for this file

ulyssessouza: Maybe a JSON linter and the same approach as we use on `V1`, which is to integrate the download of the spec file in the build.

ulyssessouza (approved):
LGTM

PR #194, opened 2021-10-11
report a nice error when service to extend is not found

ndeloof:
close https://github.com/docker/compose/issues/8762

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -450,7 +450,12 @@ func loadServiceWithExtends(filename, name string, servicesDict map[string]inter
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


loader/loader.go
```diff
@@ -450,7 +450,12 @@ func loadServiceWithExtends(filename, name string, servicesDict map[string]inter
 		return nil, err
 	}
 
-	serviceConfig, err := LoadService(name, servicesDict[name].(map[string]interface{}), workingDir, lookupEnv, opts.ResolvePaths)
+	target, ok := servicesDict[name]
+	if !ok {
+		return nil, fmt.Errorf("cannot extend service %q in %s: Service not found", name, filename)
```
mat007, line 455: nit:
```suggestion
		return nil, fmt.Errorf("cannot extend service %q in %s: service not found", name, filename)
```
?

PR #197, opened 2021-10-24
don't truncate project name computed by parent folder name with dots

ndeloof:
when project name includes some unsupported characters (like `my.project`) we must ignore them, not truncate the computed project name => `myproject`

fix https://github.com/docker/compose/issues/8822

Also introduce a dedicated `normalizeName` func to manage project name computation, vs regexp which tends to become complex to maintain :P

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -333,7 +333,7 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		} else if nameFromEnv, ok := options.Environment[ComposeProjectName]; ok && nameFromEnv != "" {
 			opts.Name = nameFromEnv
 		} else {
-			opts.Name = filepath.Base(absWorkingDir)
+			opts.Name = strings.ToLower(filepath.Base(absWorkingDir))
 		}
 		opts.Name = normalizeName(opts.Name)
 	}
--- a/cli/options.go
+++ b/cli/options.go
@@ -333,7 +333,7 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		} else if nameFromEnv, ok := options.Environment[ComposeProjectName]; ok && nameFromEnv != "" {
 			opts.Name = nameFromEnv
 		} else {
-			opts.Name = strings.ToLower(filepath.Base(absWorkingDir))
+			opts.Name = filepath.Base(absWorkingDir)
 		}
 		opts.Name = normalizeName(opts.Name)
 	}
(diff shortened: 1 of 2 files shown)


rfay:
Just a note: This isn't just a problem for projects computed from project name, it's also a problem for project names provided by `docker-compose -p` or `COMPOSE_PROJECT_NAME`. If any of those ways have '.' in them, the project name is truncated.

ndeloof:
@ulyssessouza any thoughts?

cli/options.go
```diff
@@ -335,7 +335,7 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		} else {
 			opts.Name = filepath.Base(absWorkingDir)
 		}
-		opts.Name = regexp.MustCompile(`(?m)[a-z0-9]+[-_a-z0-9]*`).FindString(strings.ToLower(opts.Name))
+		opts.Name = normalizeName(opts.Name)
```
mat007, line 338: I think the `strings.ToLower` has been removed, is it intentional?

ulyssessouza (changes requested):
As @mat007 mentioned, it's missing the `strings.ToLower`. That's what V1 does

cli/options.go
ndeloof: indeed, good catch

mat007: Don’t we need it in the `nameFromEnv` case as well? Wouldn’t it be safer to add it to `normalizeName`?

cli/options.go
```diff
@@ -333,9 +333,9 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		} else if nameFromEnv, ok := options.Environment[ComposeProjectName]; ok && nameFromEnv != "" {
 			opts.Name = nameFromEnv
 		} else {
-			opts.Name = filepath.Base(absWorkingDir)
+			opts.Name = strings.ToLower(filepath.Base(absWorkingDir))
```
ulyssessouza, line 336: What if the name comes from the environment?
I think the `strings.ToLower` should be applied inside `normalizeName` so that it doesn't depend from where it comes.

cli/options.go
ndeloof: imho `nameFromEnv` should only apply to the "name from path" scenario (but this would require some significant refactoring to existing tests)
Setting project name by .env or command line flag, I would not expect the tooling to do any magic with it, and better fail with error if I did anything wrong

cli/options.go
ndeloof: right. I prefer less magic with a name I've set explicitly, but let's go this way

ulyssessouza:
@ndeloof DCO 🤖...

It would be also nice to have at least one letter of the project name in the test to be an uppercase letter. So that would reflect the `strings.ToLower` behaviour.

rfay:
Do I need to open a separate issue for the case where the project name is not computed by parent folder name?

ulyssessouza (approved):
LGTM

ulyssessouza:
@rfay This function is meat to apply on project name regardless of its origin

PR #198, opened 2021-10-25
 Fixing variable substitution when using compose v2.

tbocek:
This fixes https://github.com/docker/compose/issues/8798, variable substitution not working. In godotenv, there seems to be an empty envMap, which should be in my opinion the lookup function, then checks the environment variables. Also to trigger this env lookup, compose-go has to call godotenv.ParseWithLookup. See also https://github.com/compose-spec/godotenv/pull/9

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/compose-spec/compose-go
 go 1.16
 
 require (
-	github.com/compose-spec/godotenv v1.0.0
+	github.com/compose-spec/godotenv v0.0.0-20211026074040-865524795bdc
 	github.com/distribution/distribution/v3 v3.0.0-20210316161203-a01c71e2477e
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.4.0
@@ -20,5 +20,3 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools/v3 v3.0.3
 )
-
-replace github.com/compose-spec/godotenv => github.com/tbocek/godotenv v1.0.1
--- a/go.mod
+++ b/go.mod
@@ -20,3 +20,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools/v3 v3.0.3
 )
+
+replace github.com/compose-spec/godotenv => github.com/tbocek/godotenv v1.0.1
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/compose-spec/compose-go
 go 1.16
 
 require (
-	github.com/compose-spec/godotenv v0.0.0-20211026074040-865524795bdc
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -20,3 +20,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools/v3 v3.0.3
 )
+
+replace github.com/compose-spec/godotenv => github.com/tbocek/godotenv v1.0.1
```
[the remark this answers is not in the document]
mat007: Now that that PR has been merged, do we want to remove that replace?

loader/loader_test.go
```diff
@@ -1811,3 +1811,24 @@ services:
 `)
 	assert.NilError(t, err)
 }
+
+func TestLoadService(t *testing.T) {
+	file, err := os.CreateTemp(os.TempDir(), "test-compose-go")
```
mat007, line 1816:
```suggestion
	file, err := os.CreateTemp("", "test-compose-go")
```
Because as the [docs](https://pkg.go.dev/os#CreateTemp) say `If dir is the empty string, CreateTemp uses the default directory for temporary files, as returned by TempDir`.

loader/loader_test.go
```diff
@@ -1811,3 +1811,24 @@ services:
...
+	assert.NilError(t, err)
+
+	m := map[string]interface{}{
+		"env_file": file.Name(),
+	}
+	s, err := LoadService("Test Name", m, ".", func(s string) (string, bool) {
+		if s == "TEST" {
+			return "YES", true
+		}
+		return "NO", false
```
mat007, line 1830: Maybe just
```suggestion
		assert.Equal(t, "TEST", s)
		return "YES", true
```
?

loader/loader_test.go
tbocek: good point!

loader/loader_test.go
tbocek: yes, makes sense.

go.mod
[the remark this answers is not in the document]
tbocek: True, I changed to: github.com/compose-spec/godotenv v0.0.0-20211026074040-865524795bdc, since there was no new tag 1.0.1 for godotenv

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/compose-spec/compose-go
 go 1.16
 
 require (
-	github.com/compose-spec/godotenv v1.0.0
+	github.com/compose-spec/godotenv v0.0.0-20211026074040-865524795bdc
```
mat007, line 6: There has just been a v1.1.0 of `compose-spec/godotenv` that’s been released. 😁

tbocek: Nice, thx, updated the version to v1.1.0

mat007 (approved):
Thanks, looks good!

PR #201, opened 2021-11-03
Fixed a typo in GroupAdd type mapping

axhav:
Found an issue that GroupAdd was not working with docker compose v2. This typo fix should restore "group_add" parameter to docker compose again.

Fixes https://github.com/docker/compose/issues/8810

Signed-off-by: William Axhav Bratt <william@axhav.se>


mat007 (approved):
Oh good catch, thanks!

mat007:
You need to sign your commit though, see https://github.com/compose-spec/compose-go/pull/201/checks?check_run_id=4091727957

axhav:
Thanks for the quick reply, signing should be fixed now

PR #203, opened 2021-11-15
Bump godotenv

ulyssessouza:
This bump fixes unquoted values on environment files.

Still depends on https://github.com/compose-spec/godotenv/pull/11 to be merged

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/compose-spec/compose-go
 go 1.16
 
 require (
-	github.com/compose-spec/godotenv v1.1.0
+	github.com/compose-spec/godotenv v1.1.1
 	github.com/distribution/distribution/v3 v3.0.0-20210316161203-a01c71e2477e
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.4.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -20,3 +20,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools/v3 v3.0.3
 )
+
+replace github.com/compose-spec/godotenv => github.com/ulyssessouza/godotenv v1.4.1-0.20211116100545-655269e25657
```
mat007, line 24: Shouldn’t we make a new release in `compose-spec/godotenv` instead?

mat007: Or at least not point to your fork, now that the upstream PR has been merged?

ndeloof: I assume this was a test PR to check the fix on godotenv side has no impact on compose-go
Maybe flag this as "do not merge" to avoid confusion?

mat007: Well, to be fair it’s in a Draft state, but I received a request for review, so I thought I would do it as if the PR were ready.

ulyssessouza:
As @ndeloof said, that's just to test. @mat007 The error was to click on "Request review"... I think my brain just went in autopilot mode and clicked. Sorry

mat007 (approved):
Yes!

PR #205, opened 2021-11-22
merge environments

ndeloof:
use MappingWithEquals#Override to handle configuration merges

close https://github.com/docker/compose/issues/8917


mat007 (approved):
Sounds good!

PR #208, opened 2021-11-24
introduce CustomLabels

ndeloof:
introduce ServiceConfig#CustomLabels to be used by compose implementation to store metadata that are **not** to be serialized as part of the compose canonical configuration.

Compose V2 can then use such labels to set `com.docker.compose.*` labels but get those excluded from the marshaled configuration or service hash. see https://github.com/docker/compose/pull/8960

see https://github.com/docker/compose/issues/8725


ulyssessouza (approved):
LGTM

PR #215, opened 2022-01-03
import godotenv as a subpackage

ndeloof:
This makes https://github.com/compose-spec/godotenv a subpackage in compose-go, as we diverged from original forked joho/godotenv to match compose requirements. This will also makes test and release process simpler

note: MIT license uses by godotenv allows relicensing into Apache. Another option would be to relicense compose-go as MIT.

see https://cloud-native.slack.com/archives/C01162AJ23C/p1639033869024300 for context


mat007:
I don’t think we can or should do it quite like that, we probably want at least to retain the original license file and copyright, adding that file to the folder should be enough: https://github.com/joho/godotenv/blob/master/LICENCE
Also what can’t re-license because we’re not the copyright holders.

ndeloof:
> we can’t re-license godotenv

we can, this is allowed by MIT license
> Permission is hereby granted to deal in the Software .. without limitation the rights to ... sublicense

but I'm fine to include the original LICENSE file in this subdirectory if you feel this is required

mat007:
>> we can’t re-license godotenv

> we can, this is allowed by MIT license

It allows anyone to sublicense, which is a little different 😬

ndeloof:
added original LICENSE as suggested

mat007 (commented):
That looks OK, this means we think we have diverged so much we’ll never be able to use upstream godotenv anyway, right?

ndeloof:
yes indeed. Also upstream project isn't very active, and initial PR https://github.com/joho/godotenv/pull/149 from @ulyssessouza (6 months old) still has no comment regarding possible approval

mat007 (approved):
All right!

It’s a bit of a shame, given how little code there is in dotenv maybe we should have written this from scratch from the beginning… 😁

ndeloof:
Agree.
Still an option if you have some spare time

ulyssessouza (approved):
LGTM

ndeloof:
I've archived compose-spec/godotenv repository so we avoid mistakes merging stuff on the wrong side

PR #218, opened 2022-01-12
Do not set undefined inherited variables

kcboschert:
Prior to Docker-Compose v2, inherited variables that were undefined on the host would not be defined in the container. This returns the original behavior.

This resolves https://github.com/docker/compose/issues/9031


flemay:
I just noticed today your PR. I created a [PR](https://github.com/compose-spec/godotenv/pull/12) with the same fix last year and created a [new one](https://github.com/compose-spec/compose-go/pull/221) in this repo.

At first, I thought the problem happened only when there were multiple variables in a `.env` file, therefore my PR was around that. Then I tried today with a single variable in `.env` file and the result confirmed the issue was still there.

You PR is more accurate!

PR #219, opened 2022-01-14
Fix default values on variables interpolation

ulyssessouza:
Handles extra closing braces by distinguishing it's chooser from the rest of the braces

Resolves https://github.com/docker/compose/issues/9024
Resolves https://github.com/docker/compose/issues/8799

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -61,12 +61,7 @@ type SubstituteFunc func(string, Mapping) (string, bool, error)
 // It accepts additional substitute function.
 func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, subsFuncs ...SubstituteFunc) (string, error) {
 	if len(subsFuncs) == 0 {
-		subsFuncs = []SubstituteFunc{
-			softDefault,
-			hardDefault,
-			requiredNonEmpty,
-			required,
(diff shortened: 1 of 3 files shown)


i-ky:
@ulyssessouza, does this fix address `:?` and `?` syntax as well? I am asking because tests only feature `:-` and `-` syntax.

ulyssessouza:
@i-ky Just added some tests for required. Thanks!

template/template.go
```diff
@@ -100,7 +103,11 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
 				if !applied {
 					continue
 				}
-				return value
+				interpolatedNested, err := SubstituteWith(rest, mapping, pattern, subsFuncs...)
+				if err != nil {
+					return ""
+				}
+				return value + interpolatedNested
```
glours, line 110: nit: 
```suggestion
				if interpolatedNested, err := SubstituteWith(rest, mapping, pattern, subsFuncs...); err == nil {
					return value + interpolatedNested
				}
				return ""
```

glours (approved):
LGTM
Only a little nit, but we can merge without fixing it

PR #226, opened 2022-02-15
delete useless code

ningmingxiao:
delete useless code


ndeloof (approved):

ndeloof:
for legal reasons, we require contributors to sign-off commits. Please amend your commit with --singoff flag, then force push your branch so we can merge

PR #231, opened 2022-02-23
Update spec and implement project name

ulyssessouza:
[Implement project name](https://github.com/compose-spec/compose-spec/pull/206)


ndeloof (approved):

johnthagen:
Did this ever get documented on the website?

- https://docs.docker.com/compose/compose-file/

I can't find any mention of the top level `name: ` attribute there.

PR #234, opened 2022-03-04
add mapping for ssh property in build section

glours:
This implement https://github.com/compose-spec/compose-spec/pull/234

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -1898,7 +1898,7 @@ services:
   test:
     build:
       context: .
-      ssh: default
+      ssh: [default]
 `)
 	assert.NilError(t, err)
 	svc, err := actual.GetService("test")
@@ -1914,7 +1914,8 @@ services:
   test:
     build:
       context: .
-      ssh: key1=value1
+      ssh: 
+        key1: value1
 `)
 	assert.NilError(t, err)
 	svc, err := actual.GetService("test")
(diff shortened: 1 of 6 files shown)


loader/loader_test.go
```diff
@@ -1867,3 +1867,59 @@ func TestLoadServiceWithVolumes(t *testing.T) {
...
+	assert.ErrorContains(t, err, "services.test.build.ssh must be a mapping")
+}
+
+func TestLoadSSHWithDefaultValueInBuildConfig(t *testing.T) {
+	actual, err := loadYAML(`
+services:
+  test:
+    build:
+      context: .
+      ssh: ssh-agent
```
ndeloof, line 1901: according to the spec proposal, should be `default`

types/types.go
```diff
@@ -426,6 +427,9 @@ func (l Labels) Add(key, value string) Labels {
 	return l
 }
 
+// SSHConfig is a mapping type for SSH build config
+type SSHConfig map[string]string
```
ndeloof, line 431: I wonder `SSHConfig` should be a `[]SSHKey`, the latter declared as struct { ID, Path }` (to match the spec)

ndeloof (approved):

ndeloof:
ready to merge, just waiting for https://github.com/compose-spec/compose-spec/pull/234 to be approved

hmajid2301:
Is this all good to merge now ? The spec has been approved ?

PR #239, opened 2022-03-16
Implement convert windows paths functionality

laurazard:
Implements [option to convert windows paths](https://docs.docker.com/compose/reference/envvars/#compose_convert_windows_paths) (was not present in v2).

Resolves docker/compose#9132

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -28,6 +28,7 @@ import (
 	"github.com/compose-spec/compose-go/errdefs"
 	"github.com/compose-spec/compose-go/loader"
 	"github.com/compose-spec/compose-go/types"
+	"github.com/compose-spec/compose-go/utils"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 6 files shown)


cli/options.go
```diff
@@ -334,6 +335,8 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 
 	options.loadOptions = append(options.loadOptions, withNamePrecedenceLoad(absWorkingDir, options))
 
+	options.loadOptions = append(options.loadOptions, withConvertWindowsPaths(options))
```
ulyssessouza, line 338: This can be added in the line above to use just one append in the section

laurazard: yeah! how does it look like now?

ulyssessouza (approved):

loader/loader.go
```diff
@@ -577,11 +579,30 @@ func LoadService(name string, serviceDict map[string]interface{}, workingDir str
 		if resolvePaths {
 			serviceConfig.Volumes[i] = resolveVolumePath(volume, workingDir, lookupEnv)
 		}
+
+		if convertPaths {
+			serviceConfig.Volumes[i] = convertVolumePath(volume)
+		}
```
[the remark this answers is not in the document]
laurazard: It wouldn’t apply it anyway as the logic in convertVolumePath returns the original in OSs different that windows anyhow, but I’m not against checking explicitly. Up to you guys.

cli/testdata/simple/compose-with-paths.yaml
```diff
@@ -0,0 +1,9 @@
+version: '3.6'
```
glours, line 1: nit: you can remove the version

glours (approved):
LGTM

PR #242, opened 2022-03-20
Fix required env vars ignored except the last one

rimelek:
During the variable substitution, if a missing variable was followed
by an existing one, the empty "err" variable overwrote the previous
non-empty variable.

Unlike with Docker Compose v1, the first found error did not guarantee
that an actual error would be thrown.

Example which would have run before the fix even though var1 is not defined:

```yaml
services:
  bash:
    image: bash:5.0.18-alpine3.15
    environment:
      var12: "_ ${var1:?Error1} _ ${var2:?Error2} _ "
    command:
      - env
```

```bash
# var2=1 docker compose up
```

This change also means that, if multiple variables are missing in one string, the first one will be reported by Docker Compose

Signed-off-by: Ákos Takács <takacs.akos@it-sziget.hu>

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -62,6 +62,7 @@ type SubstituteFunc func(string, Mapping) (string, bool, error)
 // It accepts additional substitute function.
 func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, subsFuncs ...SubstituteFunc) (string, error) {
 	var outerErr error
+	var returnErr error
 
 	result := pattern.ReplaceAllStringFunc(template, func(substring string) string {
 		_, subsFunc := getSubstitutionFunctionForTemplate(substring)
@@ -91,6 +92,9 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
 
 		if substitution == "" {
 			outerErr = &InvalidTemplateError{Template: template}
+			if returnErr == nil {
+				returnErr = outerErr
+			}
 			return ""
 		}
 
@@ -101,6 +105,9 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
 			)
 			value, applied, outerErr = subsFunc(substitution, mapping)
 			if outerErr != nil {
+				if returnErr == nil {
+					returnErr = outerErr
+				}
 				return ""
 			}
 			if applied {
@@ -119,7 +126,7 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
 		return value
 	})
 
-	return result, outerErr
+	return result, returnErr
 }
 
 func getSubstitutionFunctionForTemplate(template string) (string, SubstituteFunc) {
--- a/interpolation/interpolation_test.go
+++ b/interpolation/interpolation_test.go
@@ -114,6 +114,7 @@ func TestValidUnexistentInterpolation(t *testing.T) {
 		{test: "{{{ ${FOO:?foo_} }}}", errMsg: "foo_"},
 		{test: "{{{ ${FOO:?foo-bar-value} }}}", errMsg: "foo-bar-value"},
 		{test: "{{{ ${FOO:?foo} ${BAR:-DEFAULT_VALUE} }}}", errMsg: "foo"},
+		{test: "${FOO:?foo} ${BAR:?bar}", errMsg: "foo"},
 		{test: "{{{ ${BAR} }}}", expected: "{{{  }}}"},
 		{test: "${FOO:?baz} }}}", errMsg: "baz"},
 		{test: "${FOO?baz} }}}", errMsg: "baz"},


template/template.go
```diff
@@ -98,13 +102,19 @@ func SubstituteWith(template string, mapping Mapping, pattern *regexp.Regexp, su
 				)
 				value, applied, err = f(substitution, mapping)
 				if err != nil {
+					if errReturn == nil {
```
ndeloof, line 105: would be probably nicer to adopt https://github.com/hashicorp/go-multierror or comparable

rimelek: True. My first attempt was to handle every error message, but I didn't want to make such a "big" change, since I thought handling only one was intentional. That would be a good feature though.

When I wanted to handle multiple error messages, I used a simple slice to store the messages, but I realized I would have had to change the return value as well which could breake something else, so I returned only the first one, then I changed it to the current version.

Is the library "go-multierror" considered stable and maintained? The last change was a year ago. Of course, if something is perfect, then no need to change it :)

ndeloof: as long as the signature isn't changed and we return `error` I guess it's fine, and we can consider this "implementation detail" we could later change if required.

rimelek: Wouldn't it be a new feature though rather than a fix? With the change I made, docker compose would work as it always worked in v1, reporting the first error. As of v2, it never worked correctly. I tried with 2.0.1 on Mac getting the same behavior. When I faced this error first, I found similar issues (https://github.com/docker/compose/issues/8510) which I could not reproduce, but that was either solved since then or it happens in a specific situation which we could not find yet. I have more ideas to improve the test cases (we can discuss it somewhere else I guess) which I didn't want to implement yet either, because that would be an other change that you may or may not want to accept, but I could not add all the test cases I wanted to, since that is not possible yet without changing an other part of the code. It is possible that improving the test cases would solve other issues as well that we could not reproduce.

I am actually working on an example project on GitHub to show how PHP Composer can be used with Docker Compose without installing composer in the aplication container and somehow I always run into bugs. One was fixed in Docker Compose 2.3.1, then I ran into this error handling issue, so I thought I try to find and fix if I can instead of suggesting to my audience to use Docker Compose v1.

Your suggestion is a good one, I lke it, but I would rather not introduce a new dependency at this point. Maybe it can be solved without any dependency. For example by concatenating the error messages and returning a new error with that message, but I somehow don't think it can be decided so quickly.

If you have any suggestion what I should add to the pull request to fix parsing required variables, I would happily learn and add it, or if you have a completely different idea to fix the original issue without merging my change, that's fine to me :), but  I think you should focus on the stability more than implementing new features like this, however, it would be a good improvement in the future. 

Please, let me now, if I can help.

ndeloof: if you want to return the first error encountered, then you can just break execution by `return nil, err` and don't need to manage this `errReturn` variable

rimelek: I can't do that, because it is inside a callback function. A parameter of `ReplaceAllStringFunc`.

```golang
result := pattern.ReplaceAllStringFunc(template, func(substring string) string {
    // I set `errReturn` here
})
// This is where I can return 
```
But yes, that was my first thought, and when I reaalized that this is not a simple loop, I tried to find a way to stop `ReplaceAllStringFunc`, but I could not find a solution. The next thing I could have done was to rewrite the whole regex matching and replace it with a for loop if I find out how to do that right in Go, so I haven't tried it.

rimelek: Hi @ndeloof! Is there anything else I can do to get it fixed? :) Should I open an issue instead or do you think you can approve the pull request when you have more time to look into the code I changed? I really open to any change that 
 results a cleaner and more acceptable code while still focuses on fixing the bug.

ulyssessouza: +1 for using `multierror` here.
Also adding a test with multiple variables failing so we could cover it.

rimelek: @ulyssessouza yes, the tests need to be improved as well. It would have been my next goal, since this is why this error was not discovered before. I just added one test with two variables to validate what I changed.

As I wrote before, I totaly agree with that multierror feature, but in my mind that is a feature, which is very much required but every feature can introduce new bugs as well. When I committed this change, I could easily try it with Docker Compose and worked. I gave a presentation on a local Docker Community meetup about helping the community by answering questions on the forum, fixing small bugs and reporting issues  and I promised to talk about it in an other video in details but by that time, I was not able to test the bugfix with Docker Compose because of new changes so I gave up. 

Since I will not try to add new features in a bugfix in case of a bug which have been there for almost a year (I don't know if it is still there, because I didn't have time to check), I leave it to you how you think it should be fixed. I would just like to recommend and also ask that you try to fix the bug separately and release a patch version before implementing new changes so you can be sure that you will have a working version without any unexpected side effect.

Thank you for your work.

rimelek: I am sorry, I forgot about this PR. Now I rebased the PR on the current master, but I had to resolve merge conflicts. I hope I did it right. Since you introduced a new variable "outerErr" I renamed my variable from "errReturn" to "returnErr" to keep the code consistent. Since the condition with the variable "err" was not returned, I did not add my lines there again. 

In the meantime I also tried some multierror solution in one of my projects based on this post: https://medium.com/a-journey-with-go/go-multiple-errors-management-a67477628cf1

Eventually I decided not to use those because it wasn't clear to me how they were more than using "append" or concatenenating strings and it would not have solved my problem. I am sure you can do it right, so I can't wait to see the result.

Thank you and sorry for the delay

ulyssessouza (approved):

PR #243, opened 2022-03-29
Add a public parsing command for SSH config short syntax

glours:
To be able to parse the short syntax of an SSH config from a CLI flag, for example with Compose the reference implementation of the spec, I propose to add a simple function to easily convert a `string` to a `SSHConfig` struct

Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -1019,14 +1019,19 @@ var transformSSHConfig TransformerFunc = func(data interface{}) (interface{}, er
 		}
 		return result, nil
 	case string:
-		if value == "" {
-			value = "default"
-		}
-		key, val := transformValueToMapEntry(value, "=", false)
-		result := []types.SSHKey{{ID: key, Path: val.(string)}}
-		return result, nil
+		return ParseShortSSHSyntax(value)
 	}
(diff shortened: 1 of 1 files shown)


loader/loader.go
```diff
@@ -1019,16 +1019,21 @@ var transformSSHConfig TransformerFunc = func(data interface{}) (interface{}, er
...
-		key, val := transformValueToMapEntry(value, "=", false)
-		result := []types.SSHKey{{ID: key, Path: val.(string)}}
-		return result, nil
+		return ParseShortSSHSyntax(value)
 	}
 	return nil, errors.Errorf("expected a sting, map or a list, got %T: %#v", data, data)
 }
 
+// ParseShortSSHSyntax parse short syntax for SSH authentications
+func ParseShortSSHSyntax(value string) (interface{}, error) {
```
ndeloof, line 1028: why not return `[]SSHKey`?

ndeloof (changes requested):

loader/loader.go
glours: yep good catch, thanks!

ndeloof (approved):

PR #245, opened 2022-04-07
make HostsLists a map[string]string to better handle merging config

ndeloof:
Also introduces `.AsList()` for backward compatibility

fixes https://github.com/docker/compose/issues/9361

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -468,7 +468,16 @@ func (s SSHKey) MarshalJSON() ([]byte, error) {
 type MappingWithColon map[string]string
 
 // HostsList is a list of colon-separated host-ip mappings
-type HostsList []string
+type HostsList map[string]string
+
+// AsList return host-ip mappings as a list of colon-separated strings
+func (h HostsList) AsList() []string {
+	l := make([]string, 0, len(h))
+	for k, v := range h {
+		l = append(l, fmt.Sprintf("%s:%s", k, v))
+	}
+	return l
+}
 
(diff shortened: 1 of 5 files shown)


types/types.go
```diff
@@ -467,7 +467,16 @@ func (s SSHKey) MarshalJSON() ([]byte, error) {
...
 
 // HostsList is a list of colon-separated host-ip mappings
-type HostsList []string
+type HostsList map[string]string
+
+// AsList return host-ip mappings as a list of colon-separated strings
+func (h HostsList) AsList() []string {
+	l := make([]string, 0, len(h))
+	for k, v := range h {
+		l = append(l, fmt.Sprintf("%s:%s", k, v))
```
glours, line 476:
```suggestion
		l = append(l, fmt.Sprintf("%s: %s", k, v))
```

glours (commented):
Sounds good to me, maybe we just miss a space when generating the mapping list

types/types.go
ndeloof: the mapping-as-string-style syntax, as illustrated by https://github.com/compose-spec/compose-go/pull/245/files#diff-e03e88bd30c556736d10afdd0ba283743f9f6146e57b31cfef159e00e178af8dL170 does not use space (similar to volumes: `source:dest`) probably as inheritance from CLI parameters where spacing would make command line parsing harder

glours (approved):
LGTM

PR #248, opened 2022-04-12
introduce ServiceVolumeConfig.String()

ndeloof:
introduce ServiceVolumeConfig.String() as reverse operation for loader.ParseVolume
this can be used to create back a docker cli/engine `volume` string to be used on `bind` API

see https://github.com/docker/compose/issues/9380


glours (approved):
LGTM

PR #236, opened 2022-03-10
add secrets property to build section

glours:
This implements https://github.com/compose-spec/compose-spec/pull/238

PR #249, opened 2022-04-13
include build secrets when validating/normalizing project resources

ndeloof:
follow up https://github.com/compose-spec/compose-go/pull/236


glours (approved):
LGTM

PR #257, opened 2022-05-03
Resolve environments case-insensitively on Windows

ikedam:
Please have a look on https://github.com/docker/compose/issues/9431 for details.

compose-go extracts environment variables to `Environment map[string]string` and reference that later with simply `Environment[varname]`.
It works case-sensitively but environment variables should be treated case-insensitively on Windows.

This change handles environment variables case-insensitively on Windows, without changing data structures on compose-go.

This change also adds exposed `utils.EnvResolver` function. I plan to use that for additional changes in https://github.com/docker/compose/ , handling option args like `--build-arg` with `build` sub command and `-e` with `run` sub command.
(like this: https://github.com/ikedam/compose/pull/1/files)


ndeloof:
Thanks for looking into this!
For legal reasons, all commits need to be signed-off. Please amend your commit and force push your branch:
`git commit --amend --signoff && git push --force ...`

types/config.go
```diff
@@ -32,7 +33,7 @@ type ConfigDetails struct {
 
 // LookupEnv provides a lookup function for environment variables
 func (cd ConfigDetails) LookupEnv(key string) (string, bool) {
-	v, ok := cd.Environment[key]
+	v, ok := utils.EnvResolver(cd.Environment)(key)
```
ndeloof, line 36: as `EnvResolver` is not used in any other place, I'd prefer we keep things simple with `EnvResolverWithCase` logic being directly included here

utils/envresolver.go
```diff
@@ -0,0 +1,73 @@
...
+	// * Return the value if its name matches with the passed name case-sensitively.
+	// * Otherwise, return the value if its lower-cased name matches lower-cased passed name.
+	//     * The value is indefinite if multiple variable matches.
+	var loweredEnvironment map[string]string
+	return func(s string) (string, bool) {
+		v, ok := environment[s]
+		if ok {
+			return v, ok
+		}
+		if loweredEnvironment == nil {
```
ndeloof, line 62: No need to allocate a full lowercase map on each lookup, could just do
```go
key := strings.ToLower(s)
for k, v := range environment {
   if key == strings.ToLower(k) {
      return v
   }
}
```

utils/envresolver_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+package utils
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func Test_EnvResolverWithCase(t *testing.T) {
```
ndeloof, line 25: 👍

ndeloof (changes requested):

ikedam:
> Thanks for looking into this! For legal reasons, all commits need to be signed-off. Please amend your commit and force push your branch: `git commit --amend --signoff && git push --force ...`

I've added signed-off line as described in https://github.com/compose-spec/compose-go/blob/master/CONTRIBUTING.md :
```
PS C:\workspace\compose-go> git log -n 1
commit 90f9d92dab00e1272c050960baeaaa8d8b15aaad (HEAD -> feature/9431_CaseInsensitiveEnv, ikedam/feature/9431_CaseInsensitiveEnv)
Author: ikedam <devld@ikedam.jp>
Date:   Tue May 3 11:58:23 2022 +0900

    Resolve environments case-insensitively on Windows

    Signed-off-by: IKEDA Yasuyuki <devld@ikedam.jp>
PS C:\workspace\compose-go>
```
But something might be wrong as this is the first time I wrote signed-off line.

Should I use the real name also in author line? Or should I add GPG signature? (git commit -S)

ikedam:
Oh, I got it. I have to pass DCO action. I'll try that.

ikedam:
Thanks for the review. I fixed problems you pointed.

ndeloof (approved):

glours (approved):
LGTM

PR #258, opened 2022-05-04
Clarify committer name must be one's real name

ikedam:
Inconsistency with signed-off comment and committer name results DCO check fail.

Not documented in https://probot.github.io/apps/dco/ but failed in https://github.com/compose-spec/compose-go/runs/6267860293.

It would be nice it was clarified in CONTRIBUTING.md.


ikedam:
I found this is a common template for docker projects.
It might not be good idea to change that in a proper project...

PR #260, opened 2022-05-17
Fix windows convert paths

ulyssessouza:


cli/testdata/simple/compose-with-paths.yaml
```diff
@@ -5,3 +5,6 @@ services:
     - type: bind
       source: C:\docker\project
       target: /test
+    - type: bind
+      source: ./relative
+      target: /test-relative
```
glours, line 10: Can we also add a test with the short syntax?

glours (approved):
LGTM

PR #262, opened 2022-05-17
resolve variables in .env file based on previously set variables

ndeloof:
variable substitution in `.env` file MUST not assume values from os.Env, but use previously registered environment variables (set either by `WithOsEnv` or by direct assignment to `projectOptions.Environment`) 

variables MUST NOT override a previously defined variable, so that one can use:
- `FOO=BAR`  set by os.env
- `QIX=${FOO}` set by .env file
- `FOO=...` set by .env file does not override value set by os.env

see https://github.com/docker/compose/issues/9442


glours (approved):
LTGM

PR #263, opened 2022-05-18
add support for secret set by environment variable

ndeloof:
implements https://github.com/compose-spec/compose-spec/pull/257

### DIFF
--- a/loader/full-example.yml
+++ b/loader/full-example.yml
@@ -405,6 +405,7 @@ configs:
     external: true
   config4:
     name: foo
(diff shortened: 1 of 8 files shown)


glours (commented):
Sounds good to me but can we add a test?

ndeloof:
updated with a test-case, + added consistency check that a `secret` should have either `file` or `environment` set if it is not `external`.

glours (approved):
LGTM
Thanks for the test 😉

PR #273, opened 2022-06-07
add services circular dependency check when parse config file

junyuanz1:
fix a bug reported in https://github.com/docker/compose/issues/9526 by adding service dependency check after reading the Docker compose configuration YAML file.

Signed-off-by: Junyuan Zheng <junyuanz@outlook.com>

### DIFF
--- a/loader/dependencies.go
+++ b/loader/dependencies.go
@@ -1,0 +1,157 @@
+/*
+   Copyright 2020 Docker Compose CLI authors
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


junyuan-z:
Hey @ndeloof, let me know what you think about this change. I understand there will be some duplicate code between this repo and the `go/compose` repo, but I still think circular dependency check should live in the repo since valid YAML configuration file is the responsibility of the `compose-go` repo.

I am pretty new to those Docker repositories, so please let me know if you want me to move this change to the `compose` repo.

To solve the duplicate code issue, I am planning to to remove the `HasCycle` check from the `go/compose` once this change is checked in.

nicksieger:
Hi @junyuan-z, thanks for the contribution, I appreciate you taking the time to look at this!

While your approach looks good on its face, I still think there's room to discuss where and what kind of dependency checking of the project there should be at load time. The spec doesn't really have anything to say about this, and the current behavior does allow cycles in the project definition; depending on what options you specify, there may or may not be an error when starting services. For an example, consider a compose file with two services `foo` and `bar` that depend on each other, what should be the behavior if we run `docker compose up --no-deps foo`? Currently, that definition is allowed, and compose will only start `foo` in that case.

PR #276, opened 2022-06-24
Fix environment variable expansion

ulyssessouza:


glours (approved):
LGTM, tested against the Compose unit & e2e tests with success!

nicksieger (approved):
Nice!

dotenv/godotenv_var_expansion_test.go
```diff
@@ -0,0 +1,154 @@
+package dotenv
```
milas, line 1: 👏 for these test cases!

dotenv/parser.go
```diff
@@ -187,6 +196,8 @@ func expandEscapes(str string) string {
 			return "\n"
 		case "r":
 			return "\r"
+		case "$":
+			return "$$"
```
milas, line 200: 🤑

milas (approved):
LGTM

PR #285, opened 2022-07-12
Add gocritic to linters

ulyssessouza:


nicksieger (approved):
LGTM

laurazard (approved):
LGTM

loader/loader.go
```diff
@@ -664,16 +673,16 @@ func resolveVolumePath(volume types.ServiceVolumeConfig, workingDir string, look
 }
 
 // TODO: make this more robust
-func expandUser(path string, lookupEnv template.Mapping) string {
-	if strings.HasPrefix(path, "~") {
+func expandUser(p string, lookupEnv template.Mapping) string {
```
glours, line 676: sorry but I definitely not a fan of changing `path` to `p` to make a linter happy, we lost to much meaning of the function argument in that case! 
Isn't possible the alias the import instead?

glours (commented):
If I understand correctly one of the rule is to not use a package name as a variable one, I'm ok with that only if we don't lose the meaning of the variable in our code base, my2cents

loader/loader.go
ulyssessouza: Sure!

ulyssessouza: Renamed the import to `paths`

loader/validate.go
```diff
@@ -38,6 +38,12 @@ func checkConsistency(project *types.Project) error {
 			}
 		}
 
+		for dependedService := range s.DependsOn {
```
nicksieger, line 41: Threw in a freebie, eh? Normally I'd say separate PR but it's a small change so 👍

nicksieger (approved):
👍 ♻️

loader/validate.go
ulyssessouza: Wow! Good catch! But I didn't add that 😮 
It's already on `master` https://github.com/compose-spec/compose-go/blob/master/loader/validate.go#L41-L45

ulyssessouza: I've just rebased and it seems to remove this section from the diff

PR #290, opened 2022-07-16
Cosmetic changes. Because "any > interface{}"

ulyssessouza:
According to golang's documentation that should be good -> https://github.com/golang/go/blob/master/src/builtin/builtin.go#L94-L9
goat-beauty

### DIFF
--- a/compatibility/allowlist.go
+++ b/compatibility/allowlist.go
@@ -45,10 +45,10 @@ func (c *AllowList) supported(attributes ...string) bool {
 	return false
 }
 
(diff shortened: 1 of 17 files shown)


laurazard (approved):

AkihiroSuda:
Not all downstream projects can upgrade Go to 1.18 immediately.

Maybe this should happen after the EOL of Go 1.17?

ulyssessouza:
Closing this for waiting on EOL of 1.17

Thanks @AkihiroSuda I didn't think on it when making the PR!

PR #291, opened 2022-07-27
Add 'default' value in case of presence of a variable

ulyssessouza:


template/template_test.go
```diff
@@ -113,6 +113,42 @@ func TestEmptyValueWithHardDefault(t *testing.T) {
 	assert.Check(t, is.Equal("ok ", result))
 }
 
+func TestPrensentValueWithUnset(t *testing.T) {
```
milas, line 116: FYI you've got a typo on "Present" here (`Prensent`) which is copy-pasted to the other test cases too 😛

template/template_test.go
```diff
@@ -148,6 +184,14 @@ func TestDefaultsWithNestedExpansion(t *testing.T) {
 			template: "ok ${BAR:-${FOO} ${FOO}}",
 			expected: "ok first first",
 		},
+		{
+			template: "ok ${BAR+$FOO}",
+			expected: "ok first",
+		},
+		{
+			template: "ok ${BAR+$FOO ${FOO:+second}}",
+			expected: "ok first second",
```
milas, line 193: Omg 😵 

(nice test case, env var substitution just makes my head hurt 😂)

milas (approved):
LGTM

glours (approved):

template/template_test.go
ulyssessouza: Ops! My bad! 
Good catch!

PR #295, opened 2022-07-29
build: ensure compatibility for all supported Go versions

milas:
Use matrix testing to test on currently supported Go versions,
including 1.19 pre-release.

The `go.mod` compatibility level is also set to 1.17. Once 1.19
is released, we should bump to 1.18, as Go always supports the
current version and one older, so we should stick with this as
well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,9 @@
 on: [push, pull_request]
 name: Continuous integration
+
+permissions:
+  contents: read
+
 jobs:
   validate:
     name: validate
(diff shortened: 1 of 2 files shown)


milas:
Comment from @AkihiroSuda on #290 made me think we should have coverage here to avoid breaking things by accident.

The test suite is currently pretty fast, but we can always cut the matrix down if running 3x3 (Go x OS) gets to be too much.

AkihiroSuda (approved):

ulyssessouza (approved):

PR #296, opened 2022-08-01
Fix Interpolation interference from outer string

ulyssessouza:
Fixes the cases where the outer string also has `DefaultValue` characters.

Resolves https://github.com/docker/compose/issues/9697


laurazard (approved):
LGTM

milas (approved):

PR #294, opened 2022-07-29
use '-' as default separator and support of compatibility mode

glours:
Change the default separator used between project name and resources to `-`. This is especially needed for container names to avoid issue with hostname support.

Also add detection of the `COMPOSE_COMPATIBILITY` env variable to support the compatibility usage of the `_` separator

Fix https://github.com/docker/compose/issues/9618

PR #297, opened 2022-08-01
Revert usage of "-" as separator for resource names

glours:
As noticed by some users in the original [PR](https://github.com/compose-spec/compose-go/pull/294/files), the changes were breaking ones and unjustified regarding the value offered in counter part.

This reverts commit 56e6c331759dd691e1e33c6b8338ceb4c0ca3c50, reversing
changes made to 0ab97a220d0895c435a88ced3026478e43185c53.


ndeloof (approved):
+1
while mix of dash and underscore can create some confusion, the benefit is poor regarding the backward compatibility impact

Jyrno42:
Also pinging here for more visibility.

I am having some trouble with the updated compose version 2.9.0 where this fix should be applied. Essentially the separator still seems to be `-` when `COMPOSE_COMPATIBILITY=1` is not set.

see more context and a repro case here https://github.com/docker/compose/pull/9700#issuecomment-1206382112

PR #298, opened 2022-08-02
marshal: correctly handle empty shell commands

milas:
It's possible to override an image's entrypoint for service
container's by setting `service.entrypoint`. (Same applies for
image's command.)

In some circumstances, it's desirable to explicitly "clear"
these out from the image. However, currently, if you do specify
`entrypoint: ''` or `entrypoint: []`, when the config is
marshaled, that gets lost because of Go default value handling.

Handling this varies slightly between YAML + JSON:
 * YAML: implement `IsZeroer` interface to specify that only
   `nil` slices should be omittable & `MarshalYAML` to ensure
   that `nil` slices marshal as `null` rather than `[]`
 * JSON: do NOT use `omitempty`, meaning that the result will
   _always_ be present in the marshaled output, unfortunately,
   but it will be correct!

This preserves compatibility and "pretty" serialization for
YAML at the expense of some slightly subtle marshalling
semantics.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -88,90 +88,102 @@ type ServiceConfig struct {
 	Name     string   `yaml:"-" json:"-"`
 	Profiles []string `mapstructure:"profiles" yaml:"profiles,omitempty" json:"profiles,omitempty"`
 
-	Build             *BuildConfig                     `yaml:",omitempty" json:"build,omitempty"`
-	BlkioConfig       *BlkioConfig                     `mapstructure:"blkio_config" yaml:",omitempty" json:"blkio_config,omitempty"`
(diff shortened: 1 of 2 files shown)


milas:
See docker/compose#9634 for more context/motivation

milas:
Reviewers, for your sanity, view with whitespace changes ignored: https://github.com/compose-spec/compose-go/pull/298/files?diff=split&w=1

(`types.go` has a bunch of fields moving around due to `gofmt` alignment changes which makes it really annoying to read)

milas:
Marked this as ready for review - no changes, but I didn't hear any major concerns out-of-band with this approach 🙃

glours (approved):
Sounds go to me, just waiting for @ndeloof review to be sure there isn't hidden cases we are missing

ndeloof (approved):

PR #299, opened 2022-08-10
Pass UID and GID to compose

dmitryuk:
A lot of users and issues (https://github.com/docker/compose/issues/2380, https://github.com/docker/compose/issues/4725) of docker-compose faced with problem: how to pass UID and GID? 
So where are some crutches like `export UID` or `UID=$UID docker-compose up`.
But finally it is possible to resolve onetime on the compose side and forever forget about it.
What do you think?
As I understood next step is to add function `WithShellVariables` to https://github.com/docker/compose/blob/ca05ffe36ea889b08a9b68af413f8945e78ed867/cmd/compose/compose.go#L236


dmitryuk:
@glours @ndeloof @nicksieger Please make a review

ndeloof (commented):
I understand this need to add support for bash builtin variables (`PWD` could be considered as well). But this should be first proposed as an improvement in the compose specification
within the https://github.com/compose-spec/compose-spec/blob/master/spec.md#interpolation section

dmitryuk:
@Shardj yes, look specification PR first https://github.com/compose-spec/compose-spec/pull/282

dmitryuk:
Closed without comments, thanks

ndeloof:
my bad, the reason and I moved branches to prepare compose-go/v2 module.
Feel free to re-create your PR. But maybe we should first discuss this as a feature request: why would you want your container to run with same user as your (client-side) user?

volei:
I would want my container to run with the same user as my host user to be able to create files on volumes as a non-root user without newly-created permissions?
Am I wrong on this thought?

PR #306, opened 2022-09-08
Fix absolute paths in secrets and inherited volumes

### LINKED ISSUE #304: Unix-like absolute file secrets are transformed as relative on Windows
Environment:
  * Windows
  * minikube
  * github.com [docker.exe](https://github.com/docker/cli.git) and [docker-compose.exe](https://github.com/docker/compose.git) from Git master built with `docker buildx bake`
  * DOCKER_HOST configured to use docker daemon from minikube (`minikube docker-env | Invoke-Expression`)
  * `minikube mount ./:/minikube-docker` mount call from `C:\Users\my_user\Testing` to allow bind mounts from `docker-compose.yaml` as `/minikube-docker/<something>`

When docker-compose.yaml contains the following snippet:

```yaml
secrets:
  my_password:
    file: /minikube-docker/config/MY_PASSWORD
```

than the file is wrongly converted as relative to `C:\Users\my_user\Testing\minikube-docker\config\MY_PASSWORD` and that is also what docker-compose sees and passes to docker daemon. This leads to error `Error response from daemon: invalid mount config for type "bind": invalid mount path: 'C:/Users/my_user/Testing/minikube-docker/config/MY_PASSWORD'`.

The same works fine on volumes, though.

I added code to `docker-compose` into `cmd/compose/compose.go` immediately after `cli.ProjectFromOptions` and it clearly shows that the secret's file is already absolute, so the problem is somewhere in the `compose-go` project. I tried to check the root cause by reading the code, but I am confused by the fact that the `filepath.IsAbs` should work and `absPath` function in `loader.go` checks `filepath.IsAbs` before making the path absolute. But because I am not a Go expert, I am curently not able to compile this on Windows and run some tests.

oldium:
Volumes detect and handle unix-like absolute paths also on Windows. Secrets are translated to bind mounts, so handle them the same way. While at it, fix also inherited absolute path volumes from extended services.

Fixes #304

### DIFF
--- a/loader/full-example.yml
+++ b/loader/full-example.yml
@@ -428,6 +428,8 @@ secrets:
     environment: BAR
     x-bar: baz
     x-foo: bar
+  secret5:
+    file: /abs/secret_data
 x-bar: baz
 x-foo: bar
 x-nested:
--- a/loader/full-struct_test.go
+++ b/loader/full-struct_test.go
@@ -574,6 +574,9 @@ func secrets(workingDir string) map[string]types.SecretConfig {
 				"x-foo": "bar",
 			},
 		},
+		"secret5": {
+			File: "/abs/secret_data",
+		},
 	}
 }
 
@@ -985,6 +988,8 @@ secrets:
     environment: BAR
     x-bar: baz
     x-foo: bar
+  secret5:
+    file: /abs/secret_data
 configs:
   config1:
     file: %s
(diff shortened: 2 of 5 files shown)


oldium:
Tested also in `github.com/docker/compose` with changed `go.mod` to replace `github.com/compose-spec/compose-go` with `github.com/oldium/compose-go` (latest `fix-abs-paths` branch commit) and issue #304 is really fixed for me.

milas (approved):
LGTM

PR #307, opened 2022-09-23
env: fix errors on valid interpolation expressions

milas:
The parser was too strict here and would reject valid constructs that used an unescaped `$` that was not part of a variable expression (and not ambiguous).

Now, only errors for unmatched braced expressions (e.g. `${FOO`) are returned, but other valid cases are ignored and the `$` will be treated literally, e.g. `a $ string` -> `a $ string`, which is the same as in POSIX.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -31,7 +31,7 @@ var substitutionNamed = "[_a-z][_a-z0-9]*"
 var substitutionBraced = "[_a-z][_a-z0-9]*(?::?[-+?](.*}|[^}]*))?"
 
 var patternString = fmt.Sprintf(
-	"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?P<braced>%s)}|(?P<invalid>))",
+	"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?:(?P<braced>%s)}|(?P<invalid>)))",
 	delimiter, delimiter, substitutionNamed, substitutionBraced,
 )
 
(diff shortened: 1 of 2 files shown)


template/template_test.go
```diff
@@ -47,10 +47,24 @@ func TestSubstituteNoMatch(t *testing.T) {
...
+		actual, err := Substitute(expected, defaultMapping)
+		assert.NilError(t, err)
+		assert.Equal(t, expected, actual)
+	}
+}
+
 func TestInvalid(t *testing.T) {
 	invalidTemplates := []string{
 		"${",
-		"$}",
```
milas, line 53: This is valid and is in the new test cases above.

```shell
❯ echo "$}"
$}
```

glours:
@milas can you sign-off your commit please 😂 ?

glours (approved):
LGTM

template/template.go
```diff
@@ -31,7 +31,7 @@ var substitutionNamed = "[_a-z][_a-z0-9]*"
 var substitutionBraced = "[_a-z][_a-z0-9]*(?::?[-+?](.*}|[^}]*))?"
 
 var patternString = fmt.Sprintf(
-	"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?P<braced>%s)}|(?P<invalid>))",
+	"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?:(?P<braced>%s)}|(?P<invalid>)))",
```
milas, line 34: Here's a playground with the expanded regex and various cases: https://regex101.com/r/H5aQT8/1

nicksieger (approved):
Pretty awesome 4-character fix PR if I do say so myself!

laurazard:
Awesome!

PR #308, opened 2022-09-23
env: fix a bunch of escaping/parsing edge cases

milas:
* Handle standard escape sequences based on XSI/XBD https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html
* Do not strip `\` for non-escape sequences (e.g. `\j` should still be `\j`, not `j`)
* Fix parsing with escaped quotes at end of quoted value
* Handle key whose name starts with `export` (require whitespace between `export` keyword when stripping)
* Fix parsing of unquoted values with embedded `#` (require a space before starting an inline comment)
* Use correct parser methods throughout all test cases
* Eliminate a lot of unused code that was not working right


dotenv/cut.go
```diff
@@ -0,0 +1,9 @@
+//go:build go1.18
```
nicksieger, line 1: Does this work for everything greater than 1.18?

dotenv/godotenv.go
```diff
@@ -131,75 +124,13 @@ func Read(filenames ...string) (map[string]string, error) {
...
-// Unmarshal reads an env file from a string, returning a map of keys and values.
-func Unmarshal(str string) (map[string]string, error) {
-	return UnmarshalBytes([]byte(str))
-}
-
-// UnmarshalBytes parses env file from byte slice of chars, returning a map of keys and values.
-func UnmarshalBytes(src []byte) (map[string]string, error) {
-	return UnmarshalBytesWithLookup(src, nil)
-}
-
```
nicksieger, line 143: Should we leave these exported methods in case someone else is using them?

dotenv/godotenv.go
```diff
@@ -131,75 +124,13 @@ func Read(filenames ...string) (map[string]string, error) {
...
-		if d, err := strconv.Atoi(v); err == nil {
-			lines = append(lines, fmt.Sprintf(`%s=%d`, k, d))
-		} else {
-			lines = append(lines, fmt.Sprintf(`%s="%s"`, k, doubleQuoteEscape(v))) // nolint // Cannot use %q here
-		}
-	}
-	sort.Strings(lines)
-	return strings.Join(lines, "\n"), nil
-}
-
```
nicksieger, line 202: Ditto for this block

dotenv/godotenv_test.go
```diff
@@ -368,16 +375,19 @@ func TestParsing(t *testing.T) {
 	parseAndCompare(t, "export\tOPTION_A=2", "OPTION_A", "2")
 	parseAndCompare(t, "  export OPTION_A=2", "OPTION_A", "2")
 	parseAndCompare(t, "\texport OPTION_A=2", "OPTION_A", "2")
+	parseAndCompare(t, `export OPTION_A="export A"`, "OPTION_A", "export A")
 
 	// it 'expands newlines in quoted strings' do
 	// expect(env('FOO="bar\nbaz"')).to eql('FOO' => "bar\nbaz")
 	parseAndCompare(t, `FOO="bar\nbaz"`, "FOO", "bar\nbaz")
+	parseAndCompare(t, `FOO=a\tb`, "FOO", `a\tb`)
+	parseAndCompare(t, `FOO="a\tb"`, "FOO", "a\tb")
```
nicksieger, line 384: Should we test `\\[bvfr]` escapes too?

nicksieger (approved):
LGTM, nice cleanup

dotenv/cut_go117.go
```diff
@@ -0,0 +1,13 @@
...
+package dotenv
+
+import "bytes"
+
+func bytesCut(s, sep []byte) (before, after []byte, found bool) {
+	if i := bytes.Index(s, sep); i >= 0 {
+		return s[:i], s[i+len(sep):], true
+	}
+	return s, nil, false
+}
```
glours, line 13: Do we really need this as Golang `1.19` has been released and Go Team only support 2 versions of Golang, we may assume that `1.18` is the default minimal version now, wdyt?
I opened a [PR](https://github.com/compose-spec/compose-go/pull/309) to update the CI workflow

dotenv/godotenv.go
glours: Agreed with Nick, we should just update their internal implementation if needed
If we want to remove them in a near future, we should mark them as `deprecated` first

glours (changes requested):
IHMO no need to support Golang `1.17` anymore and to avoid breaking changes in downstream repositories we should not remove exported methods.

dotenv/godotenv.go
milas: Ahhh yeah, went overboard here, will restore these!

(FYI The only reason I started deleting code is that I found out there were actually 2x very similar but different parsing code paths! And the tests were focused on one while the actual impl code was using the other 🙈 )

milas: Re-introduced in c23e7ac3dbcfd58fab23a9666792face547d44c2!

I didn't deprecate these as they're not really problematic, I just went overboard by accident 😅 

(Not the same story for `Marshal` / `Write` tho!)

dotenv/godotenv.go
milas: Re-introduced in c23e7ac3dbcfd58fab23a9666792face547d44c2.

They're now marked Deprecated because they're really untested and divergent with how we use the library.

dotenv/godotenv_test.go
milas: Added a mega escape case in f3671f66ef9965a6350c4ce2227d31cae07352b1.

dotenv/cut_go117.go
milas: This is now gone and using `bytes.Cut` directly after merging in #309. 🙂

dotenv/cut.go
milas: Yep! Weird syntax but it's a minimum version opt-in. And thankfully it's gone now as of 3d146c71c3e41caafb84f23e3e84f77e4d83edd2!

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/compose-spec/compose-go
 
-go 1.17
+go 1.18
```
glours, line 3: 🫣 Ho I didn't check we were still using 1.17 here 😬

glours (approved):
LGTM

PR #315, opened 2022-10-25
load from file if ConfigFile.Content is empty

### LINKED ISSUE #314: Does not load from file with no content bytes.
The [godoc for `types.ConfigFile`](https://pkg.go.dev/github.com/compose-spec/compose-go@v1.6.0/types#ConfigFile) indicates that the `Content` bytes will be loaded from `Filename` if they are not set:

```go
type ConfigFile struct {
	// Filename is the name of the yaml configuration file
	Filename [string](https://pkg.go.dev/builtin#string)
	// Content is the raw yaml content. Will be loaded from Filename if not set
	Content [][byte](https://pkg.go.dev/builtin#byte)
	// Config if the yaml tree for this config file. Will be parsed from Content if not set
	Config map[[string](https://pkg.go.dev/builtin#string)]interface{}
}
```

This is not the case: https://github.com/compose-spec/compose-go/blob/v1.6.0/loader/loader.go#L169-L173

```go
		if configDict == nil {
                         // should load here if len(file.Content) == 0
			dict, err := parseConfig(file.Content, opts)
			if err != nil {
				return nil, err
			}
			configDict = dict
			file.Config = dict
			configDetails.ConfigFiles[i] = file
		}
```

flowchartsman:
- fixes #314


milas (commented):
Thanks for the PR!

Can you rebase/amend your commit to include the DCO?

See https://github.com/compose-spec/compose-go/pull/315/checks?check_run_id=9085322040 for details

flowchartsman:
Good to go (I think)

flowchartsman:
Now that is one strict DCO rule. Should be good now?

milas (approved):

laurazard (approved):
LGTM 🥳

PR #316, opened 2022-11-01
Check service secrets existence

glours:
[Regarding the Compose Specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md#secrets), a secret used at the service level  should exist on the platform or be defined in the secrets section
>Compose implementations MUST report an error if the secret doesn't exist on the platform or isn't defined in the [secrets](https://github.com/compose-spec/compose-spec/blob/master/spec.md#secrets-top-level-element) section of this Compose file.

Also fix bugs in `WithoutUnnecessaryResources` that created empty resources when a resource declared in service section weren’t existing in the top level declaration (secrets. volumes, networks, configs...)

reference: https://github.com/docker/compose/issues/9963

### DIFF
--- a/loader/validate.go
+++ b/loader/validate.go
@@ -70,6 +70,12 @@ func checkConsistency(project *types.Project) error {
 				return errors.Wrap(errdefs.ErrInvalid, fmt.Sprintf("service %q refers to undefined config %s", s.Name, config.Source))
 			}
 		}
+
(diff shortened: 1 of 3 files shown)


nicksieger (commented):
Changes look good, maybe add a couple unit tests in validate_test.go?

laurazard (approved):
LGTM 🥳

PR #320, opened 2022-11-23
Remove compatibility package

ndeloof:
Initially designed to let ECS cloud integration detect incompatible attributes, this has been being poorly maintained


milas (approved):
🧹

glours (approved):

PR #321, opened 2022-11-24
merge exposes, heathchecks.test

ndeloof:
closes https://github.com/docker/compose/issues/9756


glours (approved):
Sounds good to me, can we also add a test to test the uniqueness of the `expose` attribute?

PR #323, opened 2022-11-30
fix network aliasses merge logic to conform to docker compose v1

ndeloof:
closes https://github.com/docker/compose/issues/9981


glours (approved):
LGTM

PR #330, opened 2022-12-20
add support for build.privileged and service.cgroup

ndeloof:


types/types.go
```diff
@@ -354,12 +356,12 @@ type ThrottleDevice struct {
 // ShellCommand is a string or list of string args.
 //
 // When marshaled to YAML, nil command fields will be omitted if `omitempty`
-// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
+// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `”`)
```
glours, line 359: @ndeloof you're IDE again?
```suggestion
// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
```

types/types.go
```diff
@@ -354,12 +356,12 @@ type ThrottleDevice struct {
...
 // will serialize to an empty array (`[]`).
 //
 // When marshaled to JSON, the `omitempty` struct must NOT be specified.
 // If the command field is nil, it will be serialized as `null`.
-// Explicitly empty commands (i.e. `[]` or `''`) will serialize to an empty
+// Explicitly empty commands (i.e. `[]` or `”`) will serialize to an empty
```
glours, line 364: same?
```suggestion
// Explicitly empty commands (i.e. `[]` or `''`) will serialize to an empty
```

glours (changes requested):

types/types.go
ndeloof: 😓 I thought I found the right option to prevent this.

glours (approved):
LGTM

PR #331, opened 2022-12-21
resolve environment incrementaly, including values loaded from previous files in the list

ndeloof:
closes https://github.com/docker/compose/issues/9980

illustration example:
```
$ cat compose.yml 
services:
  test:
    image: alpine
    command: sh -c "env"
    env_file:
      - "base.env"
      - "extra.env"
$  cat base.env 
FOO="expected"
$  cat extra.env 
FOO="${FOO:-default}"
$  docker compose run test | grep FOO
FOO=expected
```

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -629,6 +629,12 @@ func convertVolumePath(volume types.ServiceVolumeConfig) types.ServiceVolumeConf
 
 func resolveEnvironment(serviceConfig *types.ServiceConfig, workingDir string, lookupEnv template.Mapping) error {
 	environment := types.MappingWithEquals{}
+	var resolve dotenv.LookupFn = func(s string) (string, bool) {
+		if v, ok := environment[s]; ok && v != nil {
+			return *v, true
+		}
+		return lookupEnv(s)
+	}
 
 	if len(serviceConfig.EnvFile) > 0 {
 		if serviceConfig.Environment == nil {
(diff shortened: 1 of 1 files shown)


types/types.go
```diff
@@ -356,12 +356,12 @@ type ThrottleDevice struct {
 // ShellCommand is a string or list of string args.
 //
 // When marshaled to YAML, nil command fields will be omitted if `omitempty`
-// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
+// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `”`)
```
laurazard, line 359: Is this change intentional?

types/types.go
```diff
@@ -356,12 +356,12 @@ type ThrottleDevice struct {
...
 // will serialize to an empty array (`[]`).
 //
 // When marshaled to JSON, the `omitempty` struct must NOT be specified.
 // If the command field is nil, it will be serialized as `null`.
-// Explicitly empty commands (i.e. `[]` or `''`) will serialize to an empty
+// Explicitly empty commands (i.e. `[]` or `”`) will serialize to an empty
```
laurazard, line 364: same here

types/types.go
ndeloof: don't know why, I get both Goland and vscode to make this change on save :'(

laurazard (approved):
LGTM 🥳

PR #332, opened 2022-12-21
resolve device path for local bind mount volumes

ndeloof:
closes https://github.com/docker/compose/issues/9971

see https://github.com/docker/compose/blob/master/compose/config/config.py#L496-L497


laurazard (approved):
Works great 🥳 LGTM

arcanisgk:
Version 2.22 on windows.
not resolved: https://github.com/docker/compose/issues/8922#issuecomment-1755793694

PR #338, opened 2023-01-05
introduce ResolveServicesEnvironment and resolve service environment AFTER profiles have been applied

ndeloof:
This introduces `WithProfiles` so the loader can directly apply selected profiles, and moves env_file loading logic to a later phase, so we won't try to load for services which have been disabled
Also introduce load option `SkipResolveEnvFiles` to fully bypass this and process `env_files` in a later phase

closes https://github.com/docker/compose/issues/8713

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -17,28 +17,31 @@
 package types
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
 
+	"github.com/compose-spec/compose-go/dotenv"
 	"github.com/distribution/distribution/v3/reference"
 	godigest "github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
 )
 
 // Project is the result of loading a set of compose files
 type Project struct {
-	Name         string            `yaml:"name,omitempty" json:"name,omitempty"`
-	WorkingDir   string            `yaml:"-" json:"-"`
-	Services     Services          `json:"services"`
-	Networks     Networks          `yaml:",omitempty" json:"networks,omitempty"`
-	Volumes      Volumes           `yaml:",omitempty" json:"volumes,omitempty"`
-	Secrets      Secrets           `yaml:",omitempty" json:"secrets,omitempty"`
-	Configs      Configs           `yaml:",omitempty" json:"configs,omitempty"`
-	Extensions   Extensions        `yaml:",inline" json:"-"` // https://github.com/golang/go/issues/6213
-	ComposeFiles []string          `yaml:"-" json:"-"`
-	Environment  map[string]string `yaml:"-" json:"-"`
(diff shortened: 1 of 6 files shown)


types/project.go
```diff
@@ -353,3 +357,49 @@ func (p *Project) ResolveImages(resolver func(named reference.Named) (godigest.D
...
+				return errors.Wrapf(err, "Failed to read %s", file)
+			}
+
+			b, err := io.ReadAll(file)
+			if err != nil {
+				return errors.Wrapf(err, "Failed to read %s", file)
+			}
+
+			// Do not defer to avoid it inside a loop
+			file.Close() //nolint:errcheck
```
milas, line 388: This will leak if we end up returning an error above - perhaps `os.ReadFile` instead?

types/types.go
```diff
@@ -357,12 +357,12 @@ type ThrottleDevice struct {
 // ShellCommand is a string or list of string args.
 //
 // When marshaled to YAML, nil command fields will be omitted if `omitempty`
-// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
+// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `”`)
```
milas, line 360: The fun quotes are back 😂

milas (commented):
A couple of type errors from GHA and one potential file leak from me but otherwise LGTM

types/project.go
glours: +1, or if we just want to pass file name we should use `envFile` variable directly

types/types.go
glours: This is one of my favorite running gags of the moment 😂

glours:
```suggestion
// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
```

types/types.go
```diff
@@ -357,12 +357,12 @@ type ThrottleDevice struct {
...
 // will serialize to an empty array (`[]`).
 //
 // When marshaled to JSON, the `omitempty` struct must NOT be specified.
 // If the command field is nil, it will be serialized as `null`.
-// Explicitly empty commands (i.e. `[]` or `''`) will serialize to an empty
+// Explicitly empty commands (i.e. `[]` or `”`) will serialize to an empty
```
glours, line 365:
```suggestion
// Explicitly empty commands (i.e. `[]` or `''`) will serialize to an empty
```

glours (changes requested):
Just few little things to fix before merging

types/types.go
ndeloof: this makes me crazy

laurazard: My LSP has turned to doing the same thing and now I have to fix it before I make my PRs 😭

glours (approved):
LGTM

milas (approved):

PR #339, opened 2023-01-05
sync compose-spec

ndeloof:
close https://github.com/docker/compose/issues/8167

also define a make target to get latest spec from reference repo

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -141,6 +141,7 @@
         },
         "cap_add": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
         "cap_drop": {"type": "array", "items": {"type": "string"}, "uniqueItems": true},
+        "cgroup": {"type": "string", "enum": ["host", "private"]},
         "cgroup_parent": {"type": "string"},
         "command": {
           "oneOf": [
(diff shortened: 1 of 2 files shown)


Lctrs:
This could be automated through Github Actions. See https://github.com/ergebnis/composer-normalize/blob/7325813b350f5208a3ee81c20b66934a58f2f2aa/.github/workflows/update.yaml for an example.

schema/compose-spec.json
```diff
@@ -581,12 +582,12 @@
       "items": {
         "type": "object",
         "properties": {
-          "capabilities": {"$ref": "#/definitions/list_of_strings"},
-          "count": {"type": ["string", "integer"]},
-          "device_ids": {"$ref": "#/definitions/list_of_strings"},
-          "driver":{"type": "string"},
-          "options":{"$ref": "#/definitions/list_or_dict"}
-        },
+            "capabilities": {"$ref": "#/definitions/list_of_strings"},
```
laurazard, line 585: are these whitespace changes intentional?

laurazard (approved):

glours (approved):

schema/compose-spec.json
ndeloof: They exist in the reference compose-spec.json file - maybe we should use a code formatter to prevent such issues

PR #341, opened 2023-01-11
automate compose-spec schema updates

ndeloof:
As suggested by https://github.com/compose-spec/compose-go/pull/339#issuecomment-1378039617
setup a cron job to sync compose-spec schema

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,29 @@
+name: Sync compose-spec
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  schema:
+    name: Update Schema
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.3.0
+
+      - name: Download schema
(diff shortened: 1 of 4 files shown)


.github/workflows/sync.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.3.0
+
+      - name: Download schema
+        run: wget --output-document=schema/compose-spec.json https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json
+
+      - name: Show diff
+        run: git diff schema/compose-spec.json
+
```
glours, line 22: Why not add those steps in the `validate` job of the CI workflow, this way having the spec in sync will be a prerequisite to merge a new PR, wdyt?

ndeloof: would work as well. No preference

glours: For me the main difference is to be sure compose-go spec version is always align with the spec everytime we merge a PR

glours (approved):
🥳 Thanks @ndeloof

PR #342, opened 2023-01-12
provide diagnostic message as we fail to parse some env file

ndeloof:
tweak original godotenv library so that we track line number and can offer relevant error message.
Also include file path we failed to parse on error


glours (approved):
Sounds good to me 👌

laurazard (approved):
nice!!

PR #346, opened 2023-01-17
fix build context path to allow url to a git repository

ezradiniz:
Related to: https://github.com/docker/compose/issues/10174


ndeloof (approved):
This is compliant with https://github.com/moby/moby/blob/master/builder/remotecontext/urlutil/urlutil.go, so LGTM

glours (approved):
LGTM

PR #349, opened 2023-01-23
don't override healthcheck test with unset test command

ndeloof:
closes https://github.com/docker/compose/issues/10188


glours (approved):
LGTM

laurazard (approved):

PR #350, opened 2023-02-15
introduce `restart` for depends_on

ndeloof:
- introduces `restart` requirement for `depends_on`
- normalize all implicit dependencies between services (shared namespace, volumes from ...) into explicit depends_on relations

This allows to select dependencies to be considered depending on the context, vs just walk the dependency graph
see https://github.com/docker/compose/pull/10284 for usage

see https://github.com/compose-spec/compose-spec/pull/305

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -256,35 +256,11 @@ const (
 
 // GetDependencies retrieve all services this service depends on
 func (s ServiceConfig) GetDependencies() []string {
-	dependencies := make(set)
-	for dependency := range s.DependsOn {
-		dependencies.append(dependency)
+	var dependencies []string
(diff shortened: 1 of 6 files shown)


types/types.go
```diff
@@ -357,12 +333,12 @@ type ThrottleDevice struct {
 // ShellCommand is a string or list of string args.
 //
 // When marshaled to YAML, nil command fields will be omitted if `omitempty`
-// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
+// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `”`)
```
laurazard, line 336: They snuck in again 😅

PR #351, opened 2023-02-15
env: fix path resolution with `extends`

milas:
* Fix an issue with `WithEnvFile` helper to not pass empty values
* Rewrite `env_file` paths from `extends` file to be relative to the included file rather than parent (this was a regression)
* Normalize `extends.file` path to be absolute when doing the same for other paths

See docker/compose#10258.

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -190,7 +190,11 @@ func WithOsEnv(o *ProjectOptions) error {
 // WithEnvFile set an alternate env file
 // deprecated - use WithEnvFiles
 func WithEnvFile(file string) ProjectOptionsFn {
-	return WithEnvFiles(file)
+	var files []string
+	if file != "" {
+		files = []string{file}
+	}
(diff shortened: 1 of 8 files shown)


cli/options.go
```diff
@@ -190,7 +190,11 @@ func WithOsEnv(o *ProjectOptions) error {
 // WithEnvFile set an alternate env file
 // deprecated - use WithEnvFiles
 func WithEnvFile(file string) ProjectOptionsFn {
-	return WithEnvFiles(file)
+	var files []string
+	if file != "" {
```
ndeloof, line 194: LGTM from a backward compatibility point of view, but we should also update compose to pass []string, see https://github.com/docker/compose/pull/10284/files#diff-2922e5cbd75acccf0c12784852f77f143247e72a57719e883225f10f960acd40L132-R232

ndeloof (approved):

PR #355, opened 2023-02-16
lint: enable linter & fix formatting issues

milas:
This also rewords some comments to avoid triggering weird smart quote behavior in `gofmt`.

PR #360, opened 2023-02-28
introduce WithDependentServices to run fn on services in reverse dependency order

ndeloof:


types/types.go
```diff
@@ -333,12 +346,12 @@ type ThrottleDevice struct {
 // ShellCommand is a string or list of string args.
 //
 // When marshaled to YAML, nil command fields will be omitted if `omitempty`
-// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `''`)
+// is specified as a struct tag. Explicitly empty commands (i.e. `[]` or `”`)
```
glours, line 349: 😂

types/types.go
```diff
@@ -333,12 +346,12 @@ type ThrottleDevice struct {
...
 // will serialize to an empty array (`[]`).
 //
 // When marshaled to JSON, the `omitempty` struct must NOT be specified.
 // If the command field is nil, it will be serialized as `null`.
-// Explicitly empty commands (i.e. `[]` or `''`) will serialize to an empty
+// Explicitly empty commands (i.e. `[]` or `”`) will serialize to an empty
```
glours, line 354: 🙈

milas: Oops, I'm fixing up #355 so we dwn't have to ever think about this again (had to deal with some Windows oddities with the linter 😒)

types/project.go
```diff
@@ -321,6 +350,35 @@ func (p *Project) ForServices(names []string) error {
 	return nil
 }
 
+// ForServicesAndDependent restrict the project model to selected services and those who depend on them
+func (p *Project) ForServicesAndDependent(names []string) error {
```
milas, line 354: I wonder if we could do this with options to `ForServices`, e.g.
```go
func (p *Project) ForServices(names []string, opt... ServiceFilterOption) error {
    ...
}

p.ForServices([]string{"foo", "bar"}, IncludeDependents)
```
(Could do the same for `WithServices` as well.)

Otherwise, small nit here: should be `ForServicesAndDependents` (plural)

types/types.go
```diff
@@ -263,6 +263,19 @@ func (s ServiceConfig) GetDependencies() []string {
 	return dependencies
 }
 
+// GetDependent retrieves all services which depend on this service
+func (s ServiceConfig) GetDependent(p *Project) []string {
```
milas, line 267: Nit:
```suggestion
func (s ServiceConfig) GetDependents(p *Project) []string {
```

Also, I wonder if passing `*Project` back in here is a bit odd (though having this next to `GetDependencies()` is nice).

Do you think `Project::GetDependentsForService(name string) []string` would make more sense?

glours (approved):

glours (approved):
Sounds good 👍

PR #153, opened 2021-07-01
adopt goyaml/v3

ndeloof:
Adopting go.yaml/v3 will bring support for yaml 1.2 schema

compatibility considerations:
- go.yaml/v3 uses 4 spaces for indentation by default. `project.MarshalYAML()` is configured to use only 2
- go.yaml/v3 force indentation on sequences (so the updated expected yaml in this PR). See https://github.com/go-yaml/yaml/pull/924

benefits: 
- v3 branch is the active development branch for go.yaml
- [nvd.nist.gov/vuln/detail/CVE-2019-11254](https://nvd.nist.gov/vuln/detail/CVE-2019-11254)
- go.yaml/v3 adds support for yaml.Node processing, which we could use to transform the model before applying to go structs. Typically can be used to introduce custom tags, like `!inline` as proposed on https://github.com/compose-spec/compose-spec/issues/298

PR #361, opened 2023-03-03
[demo] adopt goyaml/v3

ndeloof:
this is for demonstration purpose only
uses goyaml/v3 `UnmarshalYAML(node *yaml.Node)` to process nodes as they are loaded and substitute variables
Same approach allow:
- transform "short syntax" into "long syntax" and get direct mapping into go structs, without the need to rely on mapstructure.
- merge of yaml.Node trees before converting into go structs, so we can manage nil overrides (and few other corner cases)

### DIFF
--- a/interpolation/interpolation.go
+++ b/interpolation/interpolation.go
@@ -17,19 +17,15 @@
 package interpolation
 
 import (
-	"os"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


milas:
Implemented in #153

PR #362, opened 2023-03-03
Describe valid project name on error

mbland:
PR #261 changed `docker compose` behavior to require normalized project names as input, instead of normalizing project names automatically. This landed in compose-spec/compose-go v1.2.5 and docker/compose v2.5.1.

However, the previous error message gave no indication why a name isn't valid.  This is surprising for users whose previously working project names no longer work with newer versions of `docker compose`.

As of compose-spec/compose-spec#314, the spec now contains the text:

> Project name MUST contain only lowercase letters, decimal digits,
> dashes, and underscores, and MUST begin with a lowercase letter or
> decimal digit.

This updated error message provides a concise version of this requirement based on regular expression character range notation.

See also:

- compose-spec/compose-spec#311
- docker/compose#9741

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -65,7 +65,8 @@ func NewProjectOptions(configs []string, opts ...ProjectOptionsFn) (*ProjectOpti
 func WithName(name string) ProjectOptionsFn {
 	return func(o *ProjectOptions) error {
 		if name != loader.NormalizeProjectName(name) {
-			return fmt.Errorf("%q is not a valid project name", name)
+			return fmt.Errorf("%q is not a valid project name: it must contain "+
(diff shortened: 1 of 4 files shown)


mbland:
FYI, I force-pushed my branch after CI results showed I hadn't synced schema/compose-spec.json and had a `go fmt` error in cli/options.go.

ndeloof (approved):

laurazard (approved):

mbland:
Just pushed another commit—I'd missed the `loader/loader_test.go` breakage. (That's what I get for running `go test ./options` instead of `go test ./...`!)

PR #364, opened 2023-03-04
Apply package name constraints to env vars, dirs

### LINKED ISSUE #363: Project name constraints only applied to `-p`, `name:` config property
The project name constraints articulated in compose-spec/compose-spec#314 are only enforced when using `docker compose -p` or the `name:` config file property (once #362 lands in `docker compose`). Non-normalized project names from `COMPOSE_PROJECT_NAME`, the project dir, or the working dir are still allowed, which `docker compose` will then normalize.

**I'm happy to implement the necessary changes** to enforce the project name constraints across all mechanisms. I'm also happy to update the documentation PR to reflect the current behavior instead, if it's desirable to leave the current behavior as is for some reason.

## Background

I noticed this while preparing the following documentation change after having compose-spec/compose-spec#314 and #362 merged today:

- mbland/docs@62d5fcdba5f6eb94117e72dadf5c39289a7f4f70

I experimented locally using all the mechanisms described in that change to confirm the current behavior.

It seems the JSON schema change from #362 will catch invalid `name:` values in the config (per the loader/loader_test.go failure in that PR) once it lands in `docker compose`. However, the `COMPOSE_PROJECT_NAME`, project dir, and current dir mechanisms would need an update.

Maybe users are less likely to encounter issues when the project name is defined via the other mechanisms. Still, it would seem more correct and less surprising to enforce the project name constraints across all the available mechanisms.

## Related issues

- docker/compose#9741
- docker/compose#9434
- docker/compose#9429
- docker/compose#8655
- docker/compose#229

mbland:
Applies the constraints articulated in compose-spec/compose-spec#314 for project names from `COMPOSE_PROJECT_NAME`, the project dir, or the working dir.

Before this change, the constraints applied only when using `docker compose -p` or the `name:` config file property. Non-normalized project names from `COMPOSE_PROJECT_NAME`, the project dir, or the working dir were still allowed, which `docker compose` would then normalize.

The constraints are enforced by:

- `cli.ProjectFromOptions()` setting `withNamePrecedenceLoad()` as an option to `loader.Load()`.
- `loader.Load()` applies `withNamePrecedenceLoad()`, which tries to get the project name from the command line options, then `COMPOSE_PROJECT_NAME`, then the project or working dir.
- `withNamePrecedenceLoad()` calls `loader.Options.SetProjectName()`, which now also saves the original name as well as the normalized name.
- `loader.Load()` then calls `loader.projectName()`, which reads the project name from the config file(s) if it's not already set.
- `loader.projectName()` then applies the new `CheckOriginalProjectNameIsNormalized()` function extracted from `cli.WithName()`.

Only a handful of test cases required updating in `cli/options_test.go`. These test cases were checking that invalid project names were rejected by `cli.WithName()`, but that `cli.ProjectFromOptions()` would normalize them. Now `cli.ProjectFromOptions()` rejects them with the same error.

There's no need for new test cases specifically for `COMPOSE_PROJECT_NAME` or directory-based project names, because the `CheckOriginalProjectNameIsNormalized()` function is applied regardless of the source. Other existing test cases are validating that `cli.ProjectFromOptions` does also use `COMPOSE_PROJECT_NAME` or directory names when necessary.

Closes #363.

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -18,7 +18,6 @@ package cli
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -64,9 +63,9 @@ func NewProjectOptions(configs []string, opts ...ProjectOptionsFn) (*ProjectOpti
 // WithName defines ProjectOptions' name
 func WithName(name string) ProjectOptionsFn {
 	return func(o *ProjectOptions) error {
-		if name != loader.NormalizeProjectName(name) {
-			return fmt.Errorf("%q is not a valid project name: it must contain "+
-				"only characters from [a-z0-9_-] and start with [a-z0-9]", name)
+		normalized := loader.NormalizeProjectName(name)
+		if err := loader.CheckOriginalProjectNameIsNormalized(name, normalized); err != nil {
+			return err
 		}
 		o.Name = name
 		return nil
--- a/cli/options_test.go
+++ b/cli/options_test.go
@@ -52,6 +52,20 @@ func TestProjectName(t *testing.T) {
 		assert.Equal(t, p.Name, "42my_project_env")
 	})
 
+	t.Run("by name must not be empty", func(t *testing.T) {
+		_, err := NewProjectOptions([]string{"testdata/simple/compose.yaml"}, WithName(""))
+		assert.ErrorContains(t, err, `project name must not be empty`)
+	})
+
+	t.Run("by name must not come from root directory", func(t *testing.T) {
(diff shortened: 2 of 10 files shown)


mbland:
I see one of the Windows assertions failed. I can tend to this by some time Monday.

Please also see the additional comment I left: https://github.com/compose-spec/compose-go/issues/363#issuecomment-1454296475

ndeloof (approved):

mbland:
@ndeloof I appreciate the approval. Anything else needed before merging? (No rush, just following up.)

laurazard (approved):
Nice catch with the JSON marshalling :)

PR #354, opened 2023-02-16
Add `additional_contexts` to `build` service config

laurazard:
Implements the changes in https://github.com/compose-spec/compose-spec/pull/307

PR #366, opened 2023-03-13
add support for additional_contexts

ndeloof:
implement https://github.com/compose-spec/compose-spec/pull/307


laurazard:
I already had one! :p https://github.com/compose-spec/compose-go/pull/354

PR #368, opened 2023-03-15
validate service.platform is declared by build.platforms

ndeloof:
This enforces consistency doing "build" as part of "up"


laurazard:
Also, should we update https://github.com/compose-spec/compose-spec/blob/master/05-services.md#platform to mention that if `build.platforms` is set then `service.platform` needs to be one of those values?

glours (approved):

Issue #277, opened 2022-06-24
Support for non-ASCII characters in environment variable names


nevergone:
Why not allow all characters in variable names when env_file is used? Why not let the system shell handle this if Docker allows it?
Related issue: https://github.com/docker/compose/issues/8862

**Docker version 20.10.17, build 100c701**

```
$ docker run --rm -e árvíztűrő-TÜKÖRFÚRÓGÉP=ÁRVÍZTŰRŐ-tükörfúrógép ubuntu env
```

output:
```
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=18f81a2d4cf5
árvíztűrő-TÜKÖRFÚRÓGÉP=ÁRVÍZTŰRŐ-tükörfúrógép
HOME=/root
```

**docker-compose version 1.29.2, build 5becea4c**

docker-compose.yml (environment):
```
services:
  cli:
    image: ubuntu
    environment:
      árvíztűrő-TÜKÖRFÚRÓGÉP: ÁRVÍZTŰRŐ-tükörfúrógép
    command: env
```

output:
```
$ docker-compose up
Creating network "ramdisk_default" with the default driver
Creating ramdisk_cli_1 ... done
Attaching to ramdisk_cli_1
cli_1  | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
cli_1  | HOSTNAME=73fcdcfb0656
cli_1  | árvíztűrő-TÜKÖRFÚRÓGÉP=ÁRVÍZTŰRŐ-tükörfúrógép
cli_1  | HOME=/root
ramdisk_cli_1 exited with code 0
```

docker-compose.yml (env_file):
```
services:
  cli:
    image: ubuntu
    env_file:
      - env.test.txt
    command: env
```

env.test.txt:
```
árvíztűrő-TÜKÖRFÚRÓGÉP=ÁRVÍZTŰRŐ-tükörfúrógép
```

output:
```
$ docker-compose up
Starting ramdisk_cli_1 ... done
Attaching to ramdisk_cli_1
cli_1  | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
cli_1  | HOSTNAME=73fcdcfb0656
cli_1  | árvíztűrő-TÜKÖRFÚRÓGÉP=ÁRVÍZTŰRŐ-tükörfúrógép
[...]

PR #378, opened 2023-03-27
env var name CAN be any unicode char, not just ASCII

ndeloof:
make godotenv use _runes_, not `byte` to process dot env text file, so we support unicode in variable names

closes https://github.com/compose-spec/compose-go/issues/277

### DIFF
--- a/dotenv/parser_test.go
+++ b/dotenv/parser_test.go
@@ -10,20 +10,21 @@ var testInput = `
 a=b
 a[1]=c
 a.propertyKey=d
+árvíztűrő-TÜKÖRFÚRÓGÉP=ÁRVÍZTŰRŐ-tükörfúrógép
 `
 
 func TestParseBytes(t *testing.T) {
 	p := newParser()
 
-	var inputBytes = []byte(testInput)
 	expectedOutput := map[string]string{
-		"a":             "b",
-		"a[1]":          "c",
-		"a.propertyKey": "d",
+		"a":                      "b",
+		"a[1]":                   "c",
+		"a.propertyKey":          "d",
+		"árvíztűrő-TÜKÖRFÚRÓGÉP": "ÁRVÍZTŰRŐ-tükörfúrógép",
 	}
 
 	out := map[string]string{}
-	err := p.parseBytes([]byte(inputBytes), out, nil)
+	err := p.parse(testInput, out, nil)
 
 	assert.NilError(t, err)
 	assert.Equal(t, len(expectedOutput), len(out))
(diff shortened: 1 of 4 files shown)


dotenv/parser_test.go
```diff
@@ -10,20 +10,21 @@ var testInput = `
...
 
-	var inputBytes = []byte(testInput)
 	expectedOutput := map[string]string{
-		"a":             "b",
-		"a[1]":          "c",
-		"a.propertyKey": "d",
+		"a":                      "b",
+		"a[1]":                   "c",
+		"a.propertyKey":          "d",
+		"árvíztűrő-TÜKÖRFÚRÓGÉP": "ÁRVÍZTŰRŐ-tükörfúrógép",
```
milas, line 23: Missed opportunity for an emoji in a test case 🤌

milas (approved):
👍🏻 

---

From [Open Group Base Spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html):
> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2017 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in [Portable Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tagtcjh_3) and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names.

N.B. "Portable character set" is a subset of ASCII-encodable characters

So, much like we also accept other characters (`.`, `-`), don't see any reason to not allow full Unicode!

dotenv/parser_test.go
milas: oh boo, we still check to see if it's considered a letter:
```
failed to read /Users/milas/dev/repro/unicode/.env: line 1: unexpected character "😂" in variable name
```

milas: but also...
```
$ docker compose run echo
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=5fadb26ba525
TERM=xterm
😂=1
HOME=/root
```

surprise! we don't validate env var names in YAML at all 🙈 
```yaml
services:
  echo:
   environment:
     - "😂=1"
   image: alpine
   command: env
```

milas:
Agreed re: aligning with `docker run`! Sorry if my spec quoting was unclear -- it says that for _interoperability_ with POSIX-y tools you should stick to ~ASCII, but that it's not prohibited from using other characters

I think it's more of a warning that lots of those tools were built in the 70s and don't have any concept of UTF-8 😬 

But regardless, totally fine -- at the end of the day, if you pass Unicode env var names to your Compose service and it can't handle that...don't do that? 🤷🏻

PR #379, opened 2023-03-28
allow override to REMOVE elements by using `x-null`

ndeloof:
Add support for special extension `x-null` to be used in override files. When set, during merge, source value is reset to empty/null.

note: we can't just use `value: null` as I expected because this is used in a few places for a distinct purpose: environment variables to inherit local environment, or build.ssh key to be propagated by ssh-agent

This is an attempt to fix https://github.com/compose-spec/compose-spec/issues/284

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -341,6 +341,9 @@ const extensions = "#extensions" // Using # prefix, we prevent risk to conflict 
 func groupXFieldsIntoExtensions(dict map[string]interface{}) map[string]interface{} {
 	extras := map[string]interface{}{}
 	for key, value := range dict {
+		if key == extensions {
+			continue
+		}
 		if strings.HasPrefix(key, "x-") {
 			extras[key] = value
 			delete(dict, key)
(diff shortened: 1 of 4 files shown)


milas:
For reference, I think I'm leaning towards (B) and dunno if I like either (A) or (C)

ndeloof:
Without a major change to the json-schema, we have to stick with `x-*` syntax. But obviously there's no requirement to use `x-nul`, could be `x-reset-value` or `x-unset`

the two examples you described (remove element from list or environment map) is not covered by this PR. We could indeed support a custom value, or more easily use a custom yaml tag:

```yml
services:
  web:
    build:
      ssh:
        !unset  default: 
    environment:
        !unset FOO: 
```

milas:
Oh! That's way better. I was looking at the YAML spec but did not realize you could use custom tags 😅

milas:
> the two examples you described (remove element from list or environment map) is not covered by this PR

Oh, I misunderstood then 😅 I was thinking (using my same example as before) that you'd be able to do:
```yaml
services:
  web:
    build:
      context: .
      ssh:
        default:
          x-null:
    environment:
      FOO:
        x-null:
```

ndeloof:
Another challenge to address is that passing `build: null` (or comparable) will fail JSON-Schema validation :'(

ndeloof:
Created https://github.com/compose-spec/compose-go/pull/380 to explore the yaml tag approach

PR #380, opened 2023-03-30
[EN-158] allow override compose.yaml to REMOVE elements

ndeloof:
this is an alternative to https://github.com/compose-spec/compose-go/pull/379

using a custom yaml tag `!reset` (final name to be debated), this let user define compose.override.yaml to REMOVE elements from original model. 
Also supports removal from maps, like `environment`, but miss support for removal from list (but feasible)

```yaml
services:
  foo:
    image: foo
    build: !reset  
    environment:
      FOO: !reset
```

### DIFF
--- a/interpolation/interpolation.go
+++ b/interpolation/interpolation.go
@@ -18,9 +18,9 @@ package interpolation
 
 import (
 	"os"
-	"strings"
 
 	"github.com/compose-spec/compose-go/template"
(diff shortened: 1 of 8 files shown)


milas (approved):
❤️ 

The `!reset` name is good with me. Other alternative I can think of would be `!default`. (I also like `!unset` but think that might be slightly inaccurate)

PR #381, opened 2023-03-31
panic calling OverrideBy on nil map

ndeloof:
see https://github.com/docker/compose/issues/10412

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -423,6 +423,9 @@ func NewMappingWithEquals(values []string) MappingWithEquals {
 
 // OverrideBy update MappingWithEquals with values from another MappingWithEquals
 func (e MappingWithEquals) OverrideBy(other MappingWithEquals) MappingWithEquals {
+	if e == nil {
+		return other
+	}
 	for k, v := range other {
 		e[k] = v
 	}


laurazard (approved):

StefanScherer (approved):

milas (changes requested):
Normally this function mutates the receiver, but it's not here. And we're not returning a copy of `other` so subsequent modifications could also modify the original collection.

```go
func (e *MappingWithEquals) OverrideBy(other MappingWithEquals) MappingWithEquals {
	if *e == nil {
		*e = make(MappingWithEquals, len(other))
	}
	for k, v := range other {
		(*e)[k] = v
	}
	return *e
}
```
(it's ugly, there might be a nicer way? Type aliases + map pointers are always annoying 😛)

Here's the test case I was using:
```go
func TestMappingWithEqualsOverrideByNil(t *testing.T) {
	m1 := NewMappingWithEquals([]string{"FOO=1"})
	var m2 MappingWithEquals
	m3 := m2.OverrideBy(m1)
	assert.Equal(t, len(m2), 1)
	assert.Equal(t, len(m3), 1)
	v := "value"
	m1["BAZ"] = &v
	m2["BAR"] = &v

	var nilStr *string
	assert.Equal(t, m1["BAR"], nilStr)
	assert.Equal(t, m1["BAZ"], &v)

	assert.Equal(t, m2["BAR"], &v)
	assert.Equal(t, m2["BAZ"], nilStr)

	assert.Equal(t, m3["BAR"], &v)
	assert.Equal(t, m3["BAZ"], nilStr)
}
```

PR #382, opened 2023-04-04
Support legacy boolean `yes|no|on|off` but warn user about incompatibility

ndeloof:


milas (approved):

loader/loader_test.go
```diff
@@ -2126,6 +2126,16 @@ services:
 	assert.ErrorContains(t, err, "services.test.build.ssh must be a mapping")
 }
 
+func TestLoadLegacyBoolean(t *testing.T) {
+	_, err := loadYAML(`
```
advanceboy, line 2130: @glours 

Since the goal is to support legacy booleans, `actual` checking should not be removed in the test code.

loader/interpolate.go
```diff
@@ -114,10 +115,14 @@ func toFloat32(value string) (interface{}, error) {
...
 func toBoolean(value string) (interface{}, error) {
 	switch strings.ToLower(value) {
-	case "y", "yes", "true", "on":
+	case "true":
 		return true, nil
-	case "n", "no", "false", "off":
+	case "false":
 		return false, nil
+	case "y", "yes", "on":
+		return true, fmt.Errorf("%q for boolean is not supported by YAML 1.2, please use `true`", value)
```
advanceboy, line 123: @ndeloof 

If the `toBoolean` function [returns err](https://github.com/compose-spec/compose-go/pull/382/files#diff-afca43eb50fd06b0f6fc2385b324e2adf3fb2b4f73915e355a9a67e000923c56R123), [`case map[string]interface{}:` of the `recursiveInterpolate` function](https://github.com/compose-spec/compose-go/blob/950902dc33e4b070107b5becff6dcb12ac790af0/interpolation/interpolation.go#L89-L91) in `interpolation/interpolation.go` will set the output to nil.

With the current structure of the code, it seems to be difficult to successfully load with warnings.

Wouldn't it be better to just fix the `case string:` in the `recursiveInterpolate` function in `interpolation/interpolation.go`, just restore the legacy boolean support, and put the warning in a separate PR?

loader/loader_test.go
advanceboy: I agree that you don't want to support legacy YAML in the future.

However, Docker Compose v2.16 (compose-go v1.12.0) did support legacy YAML. If you are going to make breaking changes, you should do it when we bump the major version using semantic versioning.

As @ndeloof initially proposed, it is a good idea to display warnings in the current major version and make updates that will stop support in the future.
This approach will help users prepare for the changes and avoid any sudden disruptions.

ndeloof:
After debate within the team, we decided to prefer backward compatibility, warning user about this misuse of yaml, and will plan removal for this legacy syntax in a future release

PR #383, opened 2023-04-04
ignore .env can't be read if not explicitly set by user

ndeloof:
fixes https://github.com/docker/compose/issues/10430

added a test case so we don't re-introduce this regression in the future

I'm a bit sad we can't (in a portable way) just check err == EISDIR after `os.ReadFile` and need to preventively check `os.Stat`.

### DIFF
--- a/cli/options_test.go
+++ b/cli/options_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"gotest.tools/v3/assert"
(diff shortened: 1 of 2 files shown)


cli/options_test.go
```diff
@@ -308,3 +309,16 @@ func TestEnvMap(t *testing.T) {
 	m = utils.GetAsEqualsMap(l)
 	assert.Equal(t, m["foo"], "bar")
 }
+
+func TestGetEnvFromFile(t *testing.T) {
+	wd := t.TempDir()
+	f := filepath.Join(wd, ".env")
+	err := os.Mkdir(f, 0700)
```
milas, line 316:
```suggestion
	err := os.Mkdir(f, 0o700)
```
to make the linter happy

milas (approved):
LGTM!

PR #385, opened 2023-04-07
expand additional context relative paths

ndeloof:
With introduction for  `additional_contexts` a user can set context to a local path, and as we support relative paths in many places it also make sense to expand relative paths for such additional contexts. Challenge here is that such a context can be a path or URL (just like context) but also builder specific string. I'm assuming here such a custom string will ALWAYS use a `<type>://` prefix as buildx does

see https://github.com/docker/compose/pull/10369

### DIFF
--- a/loader/normalize.go
+++ b/loader/normalize.go
@@ -75,14 +75,23 @@ func Normalize(project *types.Project, resolvePaths bool) error {
 			if s.Build.Dockerfile == "" && s.Build.DockerfileInline == "" {
 				s.Build.Dockerfile = "Dockerfile"
 			}
-			localContext := absPath(project.WorkingDir, s.Build.Context)
-			if _, err := os.Stat(localContext); err == nil {
-				if resolvePaths {
+			if resolvePaths {
+				// Build context might be a remote http/git context. Unfortunately supported "remote"
+				// syntax is highly ambiguous in moby/moby and not defined by compose-spec,
+				// so let's assume runtime will check
+				localContext := absPath(project.WorkingDir, s.Build.Context)
+				if _, err := os.Stat(localContext); err == nil {
 					s.Build.Context = localContext
 				}
-				// } else {
-				// might be a remote http/git context. Unfortunately supported "remote" syntax is highly ambiguous
-				// in moby/moby and not defined by compose-spec, so let's assume runtime will check
+				for name, path := range s.Build.AdditionalContexts {
+					if strings.Contains(path, "://") { // `docker-image://` or any builder specific context type
+						continue
+					}
+					path = absPath(project.WorkingDir, path)
+					if _, err := os.Stat(path); err == nil {
+						s.Build.AdditionalContexts[name] = path
+					}
+				}
 			}
 			s.Build.Args = s.Build.Args.Resolve(fn)
 		}
--- a/types/types.go
+++ b/types/types.go
@@ -303,7 +303,7 @@ type BuildConfig struct {
 	CacheFrom          StringList            `mapstructure:"cache_from" yaml:"cache_from,omitempty" json:"cache_from,omitempty"`
 	CacheTo            StringList            `mapstructure:"cache_to" yaml:"cache_to,omitempty" json:"cache_to,omitempty"`
 	NoCache            bool                  `mapstructure:"no_cache" yaml:"no_cache,omitempty" json:"no_cache,omitempty"`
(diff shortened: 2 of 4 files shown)


ndeloof:
@crazymax does it sound reasonable to assume a custom context type will always include some `xxx://` prefix ?

crazy-max:
> @crazy-max does it sound reasonable to assume a custom context type will always include some `xxx://` prefix ?

Named context target can also be a path. Such as with buildx:

```console
docker buildx build --build-context project=path/to/project/source .
```

Otherwise yes it's either `docker-image://`, `oci-layout://`, `http://` `https://`, or Git URL.

More info: https://docs.docker.com/engine/reference/commandline/buildx_build/#build-context

ndeloof:
yes, that's my point: when not prefix is given, we can check this is a path or git url (just like for `context` attribute)
My point here is to ensure we don't introduce in the future some new context types without a prefix, which might confuse compose file parsing. Typically, being able to distinguish path vs git url is already somehow acrobatic

types/types.go
```diff
@@ -303,7 +303,7 @@ type BuildConfig struct {
 	CacheFrom          StringList            `mapstructure:"cache_from" yaml:"cache_from,omitempty" json:"cache_from,omitempty"`
 	CacheTo            StringList            `mapstructure:"cache_to" yaml:"cache_to,omitempty" json:"cache_to,omitempty"`
 	NoCache            bool                  `mapstructure:"no_cache" yaml:"no_cache,omitempty" json:"no_cache,omitempty"`
-	AdditionalContexts MappingWithEquals     `mapstructure:"additional_contexts" yaml:"additional_contexts,omitempty" json:"additional_contexts,omitempty"`
+	AdditionalContexts Mapping               `mapstructure:"additional_contexts" yaml:"additional_contexts,omitempty" json:"additional_contexts,omitempty"`
```
milas, line 306: Spec says `additional_contexts` can be sequence as well: https://github.com/compose-spec/compose-spec/blob/f5002178151d5b15f84da0deecb66a287afde19c/build.md#additional_contexts

loader/normalize.go
```diff
@@ -75,14 +75,23 @@ func Normalize(project *types.Project, resolvePaths bool) error {
...
-				// might be a remote http/git context. Unfortunately supported "remote" syntax is highly ambiguous
-				// in moby/moby and not defined by compose-spec, so let's assume runtime will check
+				for name, path := range s.Build.AdditionalContexts {
+					if strings.Contains(path, "://") { // `docker-image://` or any builder specific context type
+						continue
+					}
+					path = absPath(project.WorkingDir, path)
+					if _, err := os.Stat(path); err == nil {
+						s.Build.AdditionalContexts[name] = path
+					}
```
milas, line 93: I don't think we should `os.Stat` here -- it _sorta_ makes sense for `localContext` because it's being used as a hack there to workaround the ambiguous remote syntax

Here, since we know this is a local path, we should just normalize & use it. As it stands, if you have `additional_contexts: {foo: './bar'}` and run `docker compose config, you'll get a different result based on whether `bar` exists.

milas (changes requested):
Change to `Mapping` from `MappingWithEquals` violates the spec

EronWright:
Thanks for fixing this issue!
https://github.com/docker/compose/issues/10448

types/types.go
ndeloof: indeed. but still it's not valid to provide a key without a value, so this isn't a MappingWithEquals.
I'll add support for converting a string list into a `Mapping`

loader/normalize.go
ndeloof: >  since we know this is a local path ..

we don't. Just like `context` this can be a git URL as well. See https://docs.docker.com/engine/reference/commandline/buildx_build/#build-context

milas: Aren't we handling that above?
```go
if strings.Contains(path, "://") { // `docker-image://` or any builder specific context type
```
Or is it possible to specify a Git URL without a protocol?? (e.g. `github.com/docker/compose.git` without `https://` or `git://` etc)

ndeloof: git URL might indeed not have a `scheme://` prefix, 
especially, we support `github.com/xx/ff`  for legacy reasons. See https://github.com/moby/buildkit/blob/master/util/gitutil/git_ref.go#L54

milas: Ahhh ok I did not know that and I'm terrified now

milas (approved):

PR #386, opened 2023-04-11
schema: fix name of `dockerfile_inline` field in YAML

milas:
Missing Go struct tag here, so the field name in YAML was actually `dockerfileinline`.

(This isn't a breaking change: it was a mistake/broken before and not even usable since the schema validation would prevent you from using the typo'd field name.)

See also:
* docker/compose#10456

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -296,7 +296,7 @@ func (s set) toSlice() []string {
 type BuildConfig struct {
 	Context            string                `yaml:",omitempty" json:"context,omitempty"`
 	Dockerfile         string                `yaml:",omitempty" json:"dockerfile,omitempty"`
-	DockerfileInline   string                `yaml:",omitempty" json:"dockerfile_inline,omitempty"`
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

DYefimov:
In my case a clean debian 11 install and compose installation instructions [here](https://docs.docker.com/engine/install/debian/): 
`sudo apt-get install docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin`
lead to reproduciblity of the bug.

It looks like the official docker ppa suffers from this regression.
Fortunately enough I have dev/pre-prod/prod mechanisms, so only the dev box is affected. Nevertheless the bug forces me to rollback the version, and that leads to an environmental inconsistencies and so on and so forth... such a nuisance.
Hot-fix would be much appreciated, thanks.

ndeloof (approved):

PR #389, opened 2023-04-14
ignore .env if not a file

Strangerxxx:
`.env` should only be loaded when an actual file

it was previously fixed by https://github.com/compose-spec/compose-go/pull/124

however it took into regression

(again) fix https://github.com/docker/compose-cli/issues/1575


ndeloof (approved):

ndeloof (commented):
see https://github.com/compose-spec/compose-go/pull/383 which already address this issue

ndeloof:
see https://github.com/compose-spec/compose-go/pull/383

PR #390, opened 2023-04-14
Revival support for legacy boolean from YAML 1.1 and earlier

advanceboy:
Revival support for legacy boolean from YAML 1.1 and earlier, as originally intended in the pull request compose-spec/compose-go#382


ndeloof (commented):
Agree backward compatibility should be preferred, but user SHOULD be warned

ndeloof:
same approach was eventually selected on https://github.com/compose-spec/compose-go/pull/382

PR #391, opened 2023-04-18
schema: add `mapstructure` struct tag to `dockerfile_inline`

laurazard:
Without the proper struct tag the loader won't populate the `dockerfile_inline` field, and so this wasn't getting converted from the dict -> `BuildConfig` struct.

Adds some tests as well.

**related**:
- https://github.com/compose-spec/compose-go/pull/386
- https://github.com/docker/compose/issues/10456

**cute animal even though it's not required**:

image

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -296,7 +296,7 @@ func (s set) toSlice() []string {
 type BuildConfig struct {
 	Context            string                `yaml:",omitempty" json:"context,omitempty"`
 	Dockerfile         string                `yaml:",omitempty" json:"dockerfile,omitempty"`
-	DockerfileInline   string                `yaml:"dockerfile_inline,omitempty" json:"dockerfile_inline,omitempty"`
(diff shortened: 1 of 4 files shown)


types/types.go
```diff
@@ -88,7 +88,7 @@ type ServiceConfig struct {
 	Name     string   `yaml:"-" json:"-"`
 	Profiles []string `mapstructure:"profiles" yaml:"profiles,omitempty" json:"profiles,omitempty"`
 
-	Build        *BuildConfig `yaml:",omitempty" json:"build,omitempty"`
+	Build        *BuildConfig `yaml:"build,omitempty" json:"build,omitempty"`
```
ndeloof, line 91: doesn't hurt but not required, is this intentional ?

ndeloof (approved):

milas (approved):
This poor field 🥲

types/types.go
laurazard: ah no, I put it there while trying to figure out what was missing and forgot it 😅 I'll remove it

ndeloof:
I created https://github.com/compose-spec/compose-go/pull/392 so we don't make such mistakes anymore

PR #392, opened 2023-04-19
use yaml annotations to configure mapstructure

ndeloof:
This comes after [`dockerfile_inline`fiasco](https://github.com/compose-spec/compose-go/pull/391) 😰
Configure mapstructure to use  `yaml` annotations so we don't have to marshaller configuration
`scale` is a corner-case as default value "1" won't be ignored by `omitempty`, so use a custom `MarshalYAML|JSON` func as a workaround


milas (approved):

laurazard (approved):
Such a QOL improvement 🥳

glours (approved):

PR #395, opened 2023-05-05
load: clarify project name validation rules in errors

milas:
This error message showed a regex previously, which has confused some users, as it's not necessarily apparent that the `a-z` was implying case sensitivity.

Updated message adapted from the release notes.

* docker/compose#10512


laurazard (approved):
LGTM

PR #409, opened 2023-05-24
normalize implicit dependencies set by service:xx references

ndeloof:
this moves code that existed in docker/compose directly into the `Normalize` phase so that we get an homogeneous compose model loaded for all commands.

Used by https://github.com/docker/compose/pull/10602


loader/normalize.go
```diff
@@ -171,6 +173,61 @@ func Normalize(project *types.Project, resolvePaths bool) error {
...
+	if strings.HasPrefix(
+		ref,
+		types.ServicePrefix,
+	) {
+		return ref[len(types.ServicePrefix):], true
+	}
+	return "", false
+}
+
+func explicitDependsOn(service *types.ServiceConfig) {
```
milas, line 187: Nit: kinda vague name, maybe `inferImplicitDependencies`?

milas (approved):
Awesome 👍🏻

glours (approved):
Great 👍  thanks @ndeloof

PR #410, opened 2023-05-24
reset MUST handle `services` as a map

ndeloof:
While `services` is a mapping in yaml, this is a slice in compose-go structs.
As a result, we need to convert the yaml paths from `services.[index]` into `service.Name` as index might not match between source and override compose files, also subject to change after sort has been applied. This latter case is demonstrated by test case using a second service `bar` that becomes `services[0]` in final result

see https://dockercommunity.slack.com/archives/C2X82D9PA/p1684916162518289?thread_ts=1684571165.209559&cid=C2X82D9PA for more context

### DIFF
--- a/loader/merge_test.go
+++ b/loader/merge_test.go
@@ -1377,6 +1377,13 @@ services:
     read_only: true
     environment:
       FOO: BAR
+    ports:
+      - "8080:80"
+  bar:
+    image: test
+    ports:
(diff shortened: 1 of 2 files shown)


glours (approved):

milas (approved):
YAML + Go 😬 

Nice fix, and 🙌🏻 for test case

PR #413, opened 2023-05-31
service can use x-* name

ndeloof:
extensions are not supported inside the `services` top-level attribute, this allows users to declare services with name starting by `x-...` Unfortunately support for this is buggy, and here is a fix + test-case :)

closes https://github.com/docker/compose/issues/8809

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -605,6 +605,7 @@ func LoadServices(filename string, servicesDict map[string]interface{}, workingD
 		for k, v := range x.(map[string]interface{}) {
 			servicesDict[k] = v
 		}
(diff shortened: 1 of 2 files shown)


milas (approved):

ndeloof:
> The question is... Do we want that to have this "ambiguity"?

A huge majority of compose users don't rely on extensions (many don't know about yaml anchors), but use a simple compose.yaml file. But within those there's a significant amount of our users who might have service named `x-foo` for internal reasons

Also, this is not a new feature, just a fix for incomplete https://github.com/compose-spec/compose-go/pull/255

PR #414, opened 2023-06-02
introduce ResolveRelativePaths as a distinct operation vs Normalization

ndeloof:
`Normalization` is used both to define implicit resources _and_ to resolve relative paths, so it brings a parameter so apply load option `ResolvePaths`
this PR introduces and explicit `ResolveRelativePaths` to manage this option as an explicit load operation
this also allows to avoid duplicating the code when loading service `extends` and resolving basefile-relative paths

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -185,6 +185,7 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 			LookupValue:     configDetails.LookupEnv,
 			TypeCastMapping: interpolateTypeCastMapping,
 		},
+		ResolvePaths: true,
 	}
 
 	for _, op := range options {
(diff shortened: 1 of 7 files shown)


milas (commented):
Looks like a small issue on Windows with escapes 🙄 but LGTM! Waaaay better separation of concerns ✨

glours (approved):
LGTM

milas (approved):

loader/loader.go
```diff
@@ -185,6 +185,7 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 			LookupValue:     configDetails.LookupEnv,
 			TypeCastMapping: interpolateTypeCastMapping,
 		},
+		ResolvePaths: true,
```
tomfeigin, line 188: @ndeloof this is a change in default behavior right? Before this was false by default.
If I explicitly set it to false through the options argument of the `Load` function the behavior would be the same as before this change right?

yankay: The 'ResolvePaths' make the unit-test fail in the nerdctl :-)
relate : https://github.com/containerd/nerdctl/pull/2338

ndeloof: @tomfeigin code that has been moved into `ResolveRelativePaths` used to be inside `Normalize` which is enabled until you explicitly set `SkipNormalization`, so this is not a behavior change. Just those two processing aspects have been separated into distinct flags.

PR #416, opened 2023-06-06
introduce include to load sub-compose projects as dependencies

ndeloof:
This adds support for `include` attribute introduced in the compose-specification by https://github.com/compose-spec/compose-spec/pull/363

This supports including compose sub-projects, comparable to language dependency management

usage:
```yaml
include:
   - ../commons/compose.yaml
   - ../another-project/compose.yaml

services:
   foo:
      depends_on:
         - imported-service
```

required projects are resolved based on their own project directory to avoid relative path hell.

### DIFF
--- a/dotenv/godotenv_test.go
+++ b/dotenv/godotenv_test.go
@@ -3,11 +3,12 @@ package dotenv
 import (
 	"bytes"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/assert"
 )
 
 var noopPresets = make(map[string]string)
@@ -15,9 +16,7 @@ var noopPresets = make(map[string]string)
 func parseAndCompare(t *testing.T, rawEnvLine string, expectedKey string, expectedValue string) {
 	t.Helper()
 	env, err := Parse(strings.NewReader(rawEnvLine))
-	if !assert.NoError(t, err) {
-		return
-	}
+	assert.NilError(t, err)
 	actualValue, ok := env[expectedKey]
 	if !ok {
 		t.Errorf("Key %q was not found in env: %v", expectedKey, env)
@@ -694,4 +693,17 @@ func TestDash(t *testing.T) {
 		"VAR.WITH.DOTS":        "dots",
 		"VAR_WITH_UNDERSCORES": "underscores",
 	}, noopPresets)
+}
+
+func TestGetEnvFromFile(t *testing.T) {
+	wd := t.TempDir()
+	f := filepath.Join(wd, ".env")
+	err := os.Mkdir(f, 0o700)
+	assert.NilError(t, err)
+
+	_, err = GetEnvFromFile(nil, wd, nil)
+	assert.NilError(t, err)
+
+	_, err = GetEnvFromFile(nil, wd, []string{f})
+	assert.Check(t, strings.HasSuffix(err.Error(), ".env is a directory"))
 }
--- a/loader/include.go
+++ b/loader/include.go
@@ -1,0 +1,120 @@
+/*
+   Copyright 2020 The Compose Specification Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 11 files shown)


loader/include.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+var transformRequireConfig TransformerFunc = func(data interface{}) (interface{}, error) {
+	switch value := data.(type) {
+	case string:
+		return map[string]interface{}{"path": value}, nil
+	case map[string]interface{}:
+		return value, nil
+	default:
+		return data, errors.Errorf("invalid type %T for `require` configuration", value)
```
milas, line 42:
```suggestion
		return data, errors.Errorf("invalid type %T for `include` configuration", value)
```

milas (approved):
LGTM

loader/include.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// LoadIncludeConfig parse the require config from raw yaml
+func LoadIncludeConfig(source []interface{}) ([]types.IncludeConfig, error) {
+	var requires []types.IncludeConfig
+	err := Transform(source, &requires)
+	return requires, err
+}
+
+var transformIncludeConfig TransformerFunc = func(data interface{}) (interface{}, error) {
```
glours, line 35: nit: can't we use generics here?

loader/loader.go
```diff
@@ -262,7 +304,7 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 	}
 
 	if opts.ResolvePaths {
-		err = ResolveRelativePaths(project)
+		err := ResolveRelativePaths(project)
 		if err != nil {
 			return nil, err
 		}
```
glours, line 310: nit: could also be apply to the following `if !opts.` blocks
```suggestion
		if err := ResolveRelativePaths(project); err != nil {
			return nil, err
		}
```

loader/testdata/compose-include.yaml
```diff
@@ -0,0 +1,6 @@
+include:
+  - compose-include.yaml
+
+services:
+  foo:
+    image: foo
```
glours, line 6: nit: new line

glours (approved):
Sounds good to me, you just need to fix the conflict before merging

loader/loader.go
ndeloof: I personally consider those to have lower readability for the tiny benefit of a single line being saved

loader/include.go
ndeloof: this signature isn't a "lack of generics": those manage the parsed yaml nodes, which are actual `interface{}`

loader/loader.go
glours: As you wish, I don't have a strong opinion about this 😉

vasilvestre:
Could it work with URL later ?

ndeloof:
@vasilvestre this is indeed something we consider for future, also see https://github.com/compose-spec/compose-spec/issues/288

VladaPetrovic:
The functionality to include the remote file would be a great addition to this feature 🚀 
It was already proposed in the original [idea](https://github.com/docker/compose/issues/318#issue-37799766)

for example:
```
include:
https://localhost:8000/some-service/compose.yaml
...
```

We could leverage this feature to re-use Compose files across teams. Currently, we’re cloning an additional repository solely to include another Compose file, which feels a bit cumbersome. 🤔

The mentioned https://github.com/compose-spec/compose-spec/issues/288 seems like a broader feature with a much larger scope. While it could also address this problem, it approaches it from a different perspective.

dotenv/godotenv_test.go
```diff
@@ -3,21 +3,20 @@ package dotenv
...
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/assert"
 )
 
 var noopPresets = make(map[string]string)
 
 func parseAndCompare(t *testing.T, rawEnvLine string, expectedKey string, expectedValue string) {
 	t.Helper()
 	env, err := Parse(strings.NewReader(rawEnvLine))
```
colliniivan948-ops, line 18: ###

PR #418, opened 2023-06-16
Add substitution with options

srebhan:
This PR introduces the `SubstituteWithOptions()` function including options for custom patterns, custom substitution functions and a **new** replacement function for allowing to customize the replacement behavior if e.g. no mapping was found.
The existing `SubstituteWith` function is switched over to the newly introduced function.

### DIFF
--- a/template/template_test.go
+++ b/template/template_test.go
@@ -381,7 +381,7 @@ func TestSubstituteWithReplacementFunc(t *testing.T) {
 	options := []Option{
 		WithReplacementFunction(func(s string, m Mapping, c *Config) (string, error) {
 			if s == "${NOTHERE}" {
-				return "", fmt.Errorf("bad choice dude: %q", s)
+				return "", fmt.Errorf("bad choice: %q", s)
 			}
 			r, err := DefaultReplacementFunc(s, m, c)
 			if err == nil && r != "" {
@@ -399,7 +399,7 @@ func TestSubstituteWithReplacementFunc(t *testing.T) {
 	assert.Check(t, is.Equal("ok foobar", result))
 
 	_, err = SubstituteWithOptions("ok ${NOTHERE}", defaultMapping, options...)
-	assert.Check(t, is.ErrorContains(err, "bad choice dude"))
+	assert.Check(t, is.ErrorContains(err, "bad choice"))
 }
 
 // TestPrecedence tests is the precedence on '-' and '?' is of the first match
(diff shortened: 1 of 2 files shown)


ndeloof:
as there's no issue associated with this PR, can you please clarify how you expect this to be used ? Any upstream project relying on this ?

srebhan:
@ndeloof we are using this library for [Telegraf](https://github.com/influxdata/telegraf) and have to reproduce a specific behavior namely to only substitute environment-variable expressions if those environment variables are defined. This requirement comes from keeping backward compatibility to not break existing configurations. The issue linked above shows how we use this new function in code.

The intension is to be able to better control the replacement process while keeping the nice shell-like interpolation functionality.

Please let me know if I should open a corresponding issue!

ndeloof:
For legal reasons we require commits to be signed-off, please amend your commit and force push

ndeloof (approved):

glours (commented):
Is it possible to add an unit test checking the correct behavior of `SubstituteWithOptions` with a replacement function, please?

template/template_test.go
```diff
@@ -377,9 +377,33 @@ func TestSubstituteWithCustomFunc(t *testing.T) {
...
+	result, err := SubstituteWithOptions("ok ${FOO}", defaultMapping, options...)
+	assert.NilError(t, err)
+	assert.Check(t, is.Equal("ok first", result))
+
+	result, err = SubstituteWithOptions("ok ${BAR}", defaultMapping, options...)
+	assert.NilError(t, err)
+	assert.Check(t, is.Equal("ok foobar", result))
+
+	_, err = SubstituteWithOptions("ok ${NOTHERE}", defaultMapping, options...)
+	assert.Check(t, is.ErrorContains(err, "bad choice dude"))
```
glours, line 402: Removing the `dude` in the error message, it doesn't provide any value to your test 😉 
```suggestion
				return "", fmt.Errorf("bad choice: %q", s)
			}
			r, err := DefaultReplacementFunc(s, m, c)
			if err == nil && r != "" {
				return r, nil
			}
			return "foobar", nil
		}),
	}
	result, err := SubstituteWithOptions("ok ${FOO}", defaultMapping, options...)
	assert.NilError(t, err)
	assert.Check(t, is.Equal("ok first", result))

	result, err = SubstituteWithOptions("ok ${BAR}", defaultMapping, options...)
	assert.NilError(t, err)
	assert.Check(t, is.Equal("ok foobar", result))

	_, err = SubstituteWithOptions("ok ${NOTHERE}", defaultMapping, options...)
	assert.Check(t, is.ErrorContains(err, "bad choice"))
```

glours (commented):
Just a small modification in your test and we're ready to merge the PR

glours (approved):

PR #421, opened 2023-06-23
improve dotenv error messages

nicks:
- print the variable name that failed to parse
- print the name of the file where the syntax error appeared

### DIFF
--- a/dotenv/parser.go
+++ b/dotenv/parser.go
@@ -123,8 +123,8 @@ loop:
 			}
 
 			return "", "", inherited, fmt.Errorf(
-				`line %d: unexpected character %q in variable name`,
-				p.line, string(rune))
+				`line %d: unexpected character %q in variable name %q`,
+				p.line, string(rune), strings.Split(src, "\n")[0])
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


nicks:
the validation failure doesn't seem related to my pr -- not sure what's going on.

nicks:
ah, i see, it's waiting on https://github.com/compose-spec/compose-go/pull/420 to be merged

dotenv/parser.go
```diff
@@ -123,8 +123,8 @@ loop:
 			}
 
 			return "", "", inherited, fmt.Errorf(
-				`line %d: unexpected character %q in variable name`,
-				p.line, string(rune))
+				`line %d: unexpected character %q in variable name %q`,
+				p.line, string(rune), strings.Split(string(src), "\n")[0])
```
ndeloof, line 127: `src` is a string, doesn't seem a cast is necessary here

ndeloof (approved):

dotenv/parser.go
nicks: good point!

glours (approved):
LGTM

PR #423, opened 2023-06-28
loader: restore implicit default of `.` for build context

milas:
In the Compose spec, as well as Compose v1.x, the `build.context` field is mandatory in the object form of a `service.build` definition.

However, in compose-go (and thus Compose v2.x), this has never been enforced, and an empty `build.context` field was implicitly set to `.` aka the project directory.

Restore that behavior pending a decision on whether we want to make the spec less restrictive here so that this is acceptable or change `compose-go` to begin emitting a warning and eventually reject this.

Note that the order of normalization and resolving paths has been switched so that normalization occurs first, and then paths can be resolved across all fields consistently. This resulted in a small test update where this was incorrect before - it was loading with path resolution enabled but then asserting it had a relative path for the env file.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -261,6 +261,13 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 		Extensions:  model.Extensions,
 	}
 
+	if !opts.SkipNormalization {
+		err = Normalize(project)
+		if err != nil {
+			return nil, err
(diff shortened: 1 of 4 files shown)


ndeloof (approved):

glours (approved):
LGTM

laurazard (approved):

PR #425, opened 2023-06-28
loader: path resolution consistency

milas:
Move the project working directory absolute path resolution to `ResolveRelativePaths` and remove redundant local volume (a rarely used way of creating bind mounts) absolute path resolution from the normalization method.

### DIFF
--- a/loader/normalize.go
+++ b/loader/normalize.go
@@ -18,7 +18,6 @@ package loader
 
 import (
 	"fmt"
-	"path/filepath"
 	"strings"
 
 	"github.com/compose-spec/compose-go/errdefs"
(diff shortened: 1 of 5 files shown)


loader/normalize.go
```diff
@@ -128,14 +120,6 @@ func Normalize(project *types.Project) error {
 		project.Services[i] = s
 	}
 
-	for name, config := range project.Volumes {
-		if config.Driver == "local" && config.DriverOpts["o"] == "bind" {
-			// This is actually a bind mount
-			config.DriverOpts["device"] = absPath(project.WorkingDir, config.DriverOpts["device"])
-			project.Volumes[name] = config
-		}
-	}
```
milas, line 137: This logic was duplicated in `ResolveRelativePaths`: https://github.com/compose-spec/compose-go/blob/169d13bec1e873d0e0448e6f97b0e90cfcb71038/loader/paths.go#L54-L60

N.B. It's actually _SLIGHTLY_ different because the version in `ResolveRelativePaths` uses `resolveMaybeUnixPath`, which is arguably more correct.

ndeloof (approved):

PR #426, opened 2023-07-01
Fix: Issue in parser termination for quoted values

rushib1:
Parser fails to properly identify termination quotes for quoted strings, In a cases where the value ends with double slash followed by quotes(\\") It is not terminating at that quote and keeps moving forward assuming the quote was escaped.

I have added a test case for the changes


ndeloof (approved):

ndeloof:
for legal reasons we require commits to be signed-off by contributors, please amend your commit with `git amend --signoff` then force-push your branch

rushib1:
I have already included my signoff in the commit, Let me know If thats not what you meant.

glours (approved):

PR #428, opened 2023-07-04
add support of required attribute of depends_on

glours:
Implementation of the [following spec proposal](https://github.com/compose-spec/compose-spec/pull/382)


glours:
Check of schema validation won't pass until the Spec PR will be merged.

ndeloof (approved):

PR #431, opened 2023-07-07
Extend default replacement function with applied flag

srebhan:
In some circumstances it is important to know if an interpolation happened or the variable was actually mapped, e.g. to solve https://github.com/influxdata/telegraf/issues/13565. Extend the `DefaultReplacementFunc` by the additional flag staying compatible to the previous API...


ndeloof (approved):

ndeloof:
Seems to me this could be also used for `docker compose config` to collect variables defined in a compose file and list them, see https://github.com/docker/compose/issues/9226

glours (approved):

PR #432, opened 2023-07-11
filter depends_on services when required is not set

glours:
If a `depends_on` is marked as non-required, we shouldn't return an error if not found

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -24,6 +24,8 @@ import (
 	"path/filepath"
 	"sort"
 
+	"github.com/compose-spec/compose-go/utils"
+
 	"github.com/compose-spec/compose-go/dotenv"
 	"github.com/distribution/distribution/v3/reference"
 	godigest "github.com/opencontainers/go-digest"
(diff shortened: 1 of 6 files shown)


types/project.go
```diff
@@ -102,10 +104,19 @@ func (p *Project) ConfigNames() []string {
 
 // GetServices retrieve services by names, or return all services if no name specified
 func (p *Project) GetServices(names ...string) (Services, error) {
+	services, servicesNotFound := p.getServicesByNames(names...)
+	if len(servicesNotFound) > 0 {
+		return services, fmt.Errorf("no such service: %s", servicesNotFound[0])
```
laurazard, line 109: nit: we should return all not found services at once, otherwise if people have multiple services missing they'll have to go

```shell
$ compose up
[no such service: foo]

<fixes config>

$ compose up
[no such service: bar]

<yells at computer and fixes things again>
```

laurazard (approved):
LGTM, just left a suggestion

types/project.go
glours: As this method is public, I didn't want to introduce a breaking change by adding the support `depends_on.required` attribute, so this implementation keep the same behaviour as the current one.
I agree we should return all the missing service at once, but I prefer to do that in a dedicated PR if this is fine for you @laurazard?

ndeloof (approved):

PR #433, opened 2023-07-12
load: resolve symlinks and make path resolution more consistent

milas:
For host paths, an attempt is now made to resolve any symlinks when resolving relative paths. A few different code paths for absolute path resolution have also been unified, so there should be more consistency here, e.g. the use of `~/foo` shell-style home-relative paths now works more places.

(The only test adjustment needed was for macOS compatibility, as it used `/tmp` as a hardcoded constant, which now gets resolved, as it's a symlink 🙃 )


milas:
Reworking this, there were some flawed assumptions, I am working my up through `ProjectOptions` currently 🙉

PR #434, opened 2023-07-13
loader: consistently resolve `build.context` path

milas:
When using `extends.file` to load a service from another Compose file, if the `build.context` field wasn't explicitly set, it would get ignored by the path resolution.

As a result, the build context would end up as the working directory of the _root_ Compose file instead of the working directory of the _loaded_ Compose file.

This was because the relative path resolution logic ignored empty `build.context` values. Unfortunately, removing that restriction is itself not sufficient, as it then attempted to verify that the local path existed in an attempt to avoid unintentionally joining the working directory with a remote context value (e.g. `/foo/https://github.com/my/repo.git`).

This is problematic because the working directory itself might be relative (rooted to an unknown location that is != `.`), so it will be resolved relative to the current process working directory, and thus fail the existence check.

In practice, this happens when using `extends.file`, where we do resolve paths but intentionally pass a potentially relative value for the working dir, thus making it unsafe to be doing real filesystem operations here.

In fact, even if a context was specified, it was possible to break this by running Compose from a different directory than the _root_ Compose file while specifying the path to it.

As there's no formal specification for determining local path vs. remote build contexts (see discussion in #385), I'm simply eliminating the existence check. This COULD mean that Compose begins to erroneously consider remote context values as local paths if builders add new unsupported syntaxes, but I think it's fair for us to be more restrictive.

Additionally, I've ensured that when path resolution is happening, it _always_ resolves the `build.context` to an absolute path for consistency. In particular, this should help make it easier to use the output of `docker compose config` from arbitrary working directories.

There's a new test that covers the `extends.file` + `build.context` behavior, and everal existing test adjustments to account for the fact that Compose was emitting relative `build.context` paths from `docker compose config` despite everything else being absolute (such as volumes).

Fixes:
 * docker/for-mac#6912

### DIFF
--- a/loader/paths.go
+++ b/loader/paths.go
@@ -68,13 +68,9 @@ func ResolveRelativePaths(project *types.Project) error {
 }
 
 func ResolveServiceRelativePaths(workingDir string, s *types.ServiceConfig) {
-	if s.Build != nil && s.Build.Context != "" && !isRemoteContext(s.Build.Context) {
-		// Build context might be a remote http/git context. Unfortunately supported "remote"
-		// syntax is highly ambiguous in moby/moby and not defined by compose-spec,
-		// so let's assume runtime will check
-		localContext := absPath(workingDir, s.Build.Context)
-		if _, err := os.Stat(localContext); err == nil {
-			s.Build.Context = localContext
+	if s.Build != nil {
+		if !isRemoteContext(s.Build.Context) {
+			s.Build.Context = absPath(workingDir, s.Build.Context)
 		}
 		for name, path := range s.Build.AdditionalContexts {
 			if strings.Contains(path, "://") { // `docker-image://` or any builder specific context type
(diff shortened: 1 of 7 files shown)


loader/paths.go
```diff
@@ -68,13 +68,9 @@ func ResolveRelativePaths(project *types.Project) error {
...
-		// syntax is highly ambiguous in moby/moby and not defined by compose-spec,
-		// so let's assume runtime will check
-		localContext := absPath(workingDir, s.Build.Context)
-		if _, err := os.Stat(localContext); err == nil {
-			s.Build.Context = localContext
+	if s.Build != nil {
+		if !isRemoteContext(s.Build.Context) {
+			s.Build.Context = absPath(workingDir, s.Build.Context)
 		}
 		for name, path := range s.Build.AdditionalContexts {
```
milas, line 75: N.B. There's actually an additional fix here for `build.additional_contexts` -- they were only being resolved if the full condition was met: `s.Build != nil && s.Build.Context != "" && !isRemoteContext(s.Build.Context)`

milas:
Ahhh I busted the tests on Windows, will look at it in the morning, at a first glance I think we need to use `resolveMaybeUnixPath` on the build context & additional contexts but need to think on it more 🤔

milas:
See also:
* docker/compose#10798

glours (approved):

PR #435, opened 2023-07-13
introduce ResourceResolver to accept remote resources

ndeloof:
`ResourceLoaders` can be configured by Compose implementation to support remote resources with `extends` and `include`. This will allow users to rely on compose.yaml files published by others, without the need for a mono-repo.

illustration example:
```yaml 
include:
   - git@github.com:acme/commons.git#v2.0:compose.yaml
```

If this feature demonstrates some significant use for a specific protocol (git, http, oci-artifacts ?) we _could_ make those part of the compose specification and offer a default loader implementation in compose-go

We also could extend this support to other places compose loads files, typically with `env_file` and `secret.file` but I'd prefer we keep this focussed on a simpler use-case for a first round.

https://docker.atlassian.net/browse/ENV-251


loader/loader.go
```diff
@@ -67,7 +67,14 @@ type Options struct {
...
 	// Profiles set profiles to enable
-	Profiles []string
+	Profiles        []string
+	ResourceLoaders []ResourceLoader
+}
+
+// ResourceLoader is a plugable remote resource resolver
+type ResourceLoader interface {
+	Accept(string) bool
+	Load(string) (string, error)
```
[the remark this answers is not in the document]
ndeloof: yes indeed sorry, my plan was to offer an illustration example, but still deserves some comments on API

loader/testdata/remote/compose.yaml
```diff
@@ -0,0 +1,3 @@
+services:
+  foo:
+    image: foo
```
laurazard, line 3: nit: end file with a newline

ndeloof:
added example code using this feature to support git resources: https://github.com/docker/compose/pull/10811

loader/paths.go
```diff
@@ -105,7 +105,7 @@ func ResolveServiceRelativePaths(workingDir string, s *types.ServiceConfig) {
 	}
 }
 
-func absPath(workingDir string, filePath string) string {
+func AbsPath(workingDir string, filePath string) string {
```
milas, line 108: Do we need to export this? This func is slightly problematic TBH, so I'd rather not expand its usage (I actually got rid of it/replaced it in #436)

loader/loader.go
```diff
@@ -68,6 +69,16 @@ type Options struct {
...
+	// ResourceLoaders manages support for remote resources
+	ResourceLoaders []ResourceLoader
+}
+
+// ResourceLoader is a plugable remote resource resolver
+type ResourceLoader interface {
+	// Accept returns `true` is the resource reference matches ResourceLoader supported protocol(s)
+	Accept(path string) bool
+	// Load returns the path to a local copy of remote resource identified by `path`.
+	Load(ctx context.Context, path string) (string, error)
```
milas, line 81: Would it make sense to actually return the contents as `[]byte` here? Internally, loader implementations could do on-disk caching for performance, but externally it could be more of a "store"/VFS interface.

That'd let us use this consistently in all code-paths, aka have a `FilesystemLoader` as the default/fallback. (Not suggesting that as part of this PR, just looking forward.)

That'd make testing easier in some cases, and hopefully help address some of the implicit direct I/O that happens, e.g. `include` / `extends` / file references in secrets, `env_file`, etc

loader/paths.go
ndeloof: AFAICT the `RealAbsPath` you introduce is a drop-in replacement for this one, isn't it ? Then could just use it

loader/loader.go
ndeloof: I had this in mind first, but there are many places we also load sibling content, and as such we really need a path so we can search within parent folder.

laurazard: Right, I had the same thought and then ended up looking around and deciding that returning a local path might make more sense in the long run. Okay with both implementations though.

glours (approved):
Sounds good to me 👍

loader/loader.go
ulyssessouza: Joining @milas on this, I would actually love to see some [Afero](https://github.com/spf13/afero) in action here. 
It looks like the [CopyOnWriteFs](https://github.com/spf13/afero#copyonwritefs) would look great for this use case.
And it would handle the case of sibling files too.

milas: Ah yeah the sibling content makes sense and needs more thought. I guess a good first step (independent of this) would be to virtualize/abstract the direct filesystem operations happening and could then offer an updated loader interface that works with that.

@ulyssessouza Good reminder about Afero, we could definitely make interfaces compatible with them in `compose-go` so it's easy to plug their implementations in. There's also the newer `io.FS` stuff in Go which might work since `compose-go` only reads but never writes.

loader/include.go
```diff
@@ -65,7 +74,7 @@ func loadInclude(configDetails types.ConfigDetails, model *types.Config, options
 			return nil, err
 		}
 
-		imported, err := load(types.ConfigDetails{
+		imported, err := load(nil, types.ConfigDetails{
```
milas, line 77:
```suggestion
		imported, err := load(ctx, types.ConfigDetails{
```

loader/loader.go
```diff
@@ -193,8 +204,13 @@ func parseYAML(source []byte) (map[string]interface{}, PostProcessor, error) {
 	return converted.(map[string]interface{}), &processor, nil
 }
 
-// Load reads a ConfigDetails and returns a fully loaded configuration
+// Load deprecated, prefer LoadWithContext
```
milas, line 207:
```suggestion
// Load reads a ConfigDetails and returns a fully loaded configuration.
//
// Deprecated: use LoadWithContext.
```

This should make it show up as deprecated in IDEs and hide it on `pkg.dev`

loader/testdata/remote/compose.yaml
milas, line 3: It'd be good to see some paths in here along with corresponding assertions.

loader/loader_test.go
```diff
@@ -2558,3 +2559,46 @@ services:
...
+	return strings.HasPrefix(s, c.prefix+":")
+}
+
+func (c customLoader) Load(ctx context.Context, s string) (string, error) {
+	path := filepath.Join("testdata", c.prefix, "compose.yaml")
+	_, err := os.Stat(path)
+	return path, err
+}
+
+func TestLoadWithRemoteResources(t *testing.T) {
```
milas, line 2577: I think we could probably use a few additional test cases, e.g. when remote loader returns an error, when you load a remote file that loads ANOTHER remote file, cycle detection with remote resources 😬

milas (commented):
As a POC, it LGTM, but this does go against the spec: https://docs.docker.com/compose/compose-file/05-services/#finding-referenced-service
> `file` value can be:
>   * Not present. This indicates that another service within the same Compose file is being referenced.
>   * File path, which can be either:
>     * Relative path. This path is considered as relative to the location of the main Compose file.
>     * Absolute path.

Everything in the spec is about **file** paths currently and currently `compose-go` is all that's needed to load any arbitrary Compose file.

With this change, if a Compose file gets loaded that uses remote resources, if the _caller_ doesn't have an appropriate loader in place, it will silently fallback to trying to make things file paths, e.g. `https://example.com/compose.yaml` -> `/Users/milas/myproject/https://example.com/compose.yaml` and then fail.

At a minimum, I think we need to define an unambiguous distinction in the spec between local file paths and remote resources, e.g. remote resource must start with a protocol (`[a-z]+://`), and if no loader is registered for the desired protocol, an error is returned. Values without a protocol (or an explicit `file://`) could continue to be treated as file paths.

`extends.file` could be the first part then updated to reference "a resource path" as a generic concept, for example.

Beyond that, I'd also argue we should consider having a standard set of remote loaders (http(s), git) live in SUBmodules of `compose-go` so that they're usable outside of `docker/compose`.

loader/loader.go
ndeloof: I agree it would be nice we could rely on some higher level abstraction like the `io.fs` package, but as this all results into a `types.ConfigDetails` which uses file paths as `string`s this won't be very relevant afaict (not to mention the required refactoring is far from being trivial)

ndeloof: @ulyssessouza @milas we can't use such an abstraction as we need an actual file path to populate attributes we pass directly to engine API, like build context or bind mount source.

PR #436, opened 2023-07-13
load: resolve symlinks and fix various path issues

milas:
The loader is very inconsistent with how it deals with relative
paths, and there have been various edge cases in which it does
not behave as expected (see #434).

At its core, this commit attempts to do the following:
  * Within the loader, do not perform real filesystem operations
    on paths (e.g. stat for existence)
  * Within the loader, use the root paths provided as-is, making
    other paths relative to them without trying to make the final
    result absolute
  * Perform real path lookup in the `ProjectOptions` helper in
    the `cli` package, including resolving symlinks and making the
    root paths absolute

Effectively, this is an attempt to make (the path-handling parts
of) the loader more "pure". It still does a lot of direct file
I/O (e.g. for `extends`, `env_file`) using `os.ReadFile` directly,
but this goes a long way towards making it practical to use the
loader for complex, multi-file projects without ever actually
touching the disk.

Pushing more of the "real system" concerns up to `ProjectOptions`
in the `cli` package means that the behavior should still be
consistent for high-level consumers.

There is a breaking behavioral change here for low-level consumers,
however, as shell-style user-home relative paths (e.g. `~/foo`)
were being resolved directly using system lookups deep within the
loader. Now, `ConfigDetails` includes a `HomeDir` field much like
`WorkingDir`, which is used to resolve these types of paths.

In the end, this should all provide greater flexibility and make
things slightly easier to test, since there's less system leakage.

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -75,6 +75,10 @@ type ProjectOptions struct {
 	// exist or an error will be returned during load.
 	EnvFiles []string
 
+	// HomeDir is the user's home directory to be used for resolving shell-style
+	// `~/foo` home-relative paths, which are supported by path fields in Compose.
+	HomeDir string
+
 	loadOptions []func(*loader.Options)
 }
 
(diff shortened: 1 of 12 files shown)


cli/options.go
```diff
@@ -358,19 +392,20 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 	if err != nil {
 		return nil, err
 	}
-	absWorkingDir, err := filepath.Abs(workingDir)
+	workingDir, err = filepath.Abs(workingDir)
```
ndeloof, line 395: why not also use RealAbsPath here ?

cli/paths.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"io/fs"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// RealAbsPath attempts to determine the true absolute path after symlink evaluation.
+func RealAbsPath(path string) (string, error) {
```
ndeloof, line 27: nit: why not just name this `AbsPath`?

ndeloof (approved):

PR #442, opened 2023-08-04
Parse service device count to int if possible

XiangRongLin:
When set via an environment variable an integer will be represented by a string, resulting in an error.
fixes: https://github.com/docker/compose/issues/10667

--- 

The alternative would be to adjust the parsing of the yaml into the internal model, so it gets parsed as an int. 
The env var is replaced in https://github.com/compose-spec/compose-go/blob/95ac1be8bfdc4dc21cdfb14e37b52c31275046a3/interpolation/interpolation.go#L74
 But it only allows setting strings and not int
https://github.com/compose-spec/compose-go/blob/95ac1be8bfdc4dc21cdfb14e37b52c31275046a3/interpolation/interpolation.go#L34

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -1038,7 +1038,12 @@ var transformServiceDeviceRequest TransformerFunc = func(data interface{}) (inte
 					value["count"] = -1
 					return value, nil
 				}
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

glours (approved):
LGTM

PR #446, opened 2023-08-21
load: move env var profile detection to option

milas:
Instead of reading `COMPOSE_PROFILES` environment variable directly during `load()`, rely exclusively on `opts.Profiles`.

A new loader option, `WithDefaultProfiles`, allows passing profile name(s) via varargs. If none are provided, it will fallback to using any profiles set via `COMPOSE_PROFILES`.

This improves purity of the loader and fixes issues with profiles being ignored for `include` (docker/compose#10906).

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -194,7 +194,7 @@ func WithEnv(env []string) ProjectOptionsFn {
 	}
 }
 
-// WithDiscardEnvFiles sets discards the `env_file` section after resolving to
+// WithDiscardEnvFile sets discards the `env_file` section after resolving to
 // the `environment` section
 func WithDiscardEnvFile(o *ProjectOptions) error {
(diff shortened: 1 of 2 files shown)


milas:
Also, technically `COMPOSE_PROFILES` is not part of the spec (we don't specify HOW things like CLI should work).

At the same time, one thing we've learned from `docker/cli` is that it's really annoying if a bunch of conventions (`DOCKER_CONTEXT`) aren't supported by the "lower-level" client (`moby/moby`). So that's why I went with the `WithDefaultProfiles` loader option here - it keeps the logic in `compose-go` for the broader ecosystem, but it's also easy to avoid by using `WithProfiles` instead.

ndeloof:
@milas loader definitively should not use os.GetEnv

ndeloof (approved):

PR #447, opened 2023-08-21
load: ignore dependency fields with `extends`

milas:
When using `extends`, clear any dependency fields on the base service before merging.

See:
 * docker/compose#10905

Draft until we determine what behavior is actually desirable.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -777,6 +777,21 @@ func loadServiceWithExtends(ctx context.Context, filename, name string, services
 			ResolveServiceRelativePaths(baseFileParent, baseService)
 		}
 
+		baseService.VolumesFrom = nil
+		baseService.DependsOn = nil
+		maybeReferences := []*string{
+			&baseService.NetworkMode,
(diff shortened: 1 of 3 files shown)


ndeloof:
:+1 This indeed need to be addressed
IMHO we should not silently ignore incompatible attributes. Either return an error (the compose v1 way) or warn user those are ignored

glours:
IHMO we should let the user extends a services with dependencies without backporting the original configuration, this way users will be able to redifine the same or new dependencies. 
An error message will blocked them from extending the service.

ndeloof:
@glours I'm worried we silently remove stuff from a service and user isn't warned, then will complain
this limitation was in place in compose v1 and AFAICT nobody complained :)

glours:
@ndeloof I was updating my message when you answered, for me we should warn the user that the original configuration won't apply in this cases

yariv-apiiro:
Perhaps we should change the docs for now? I personally uses it as well.

PR #449, opened 2023-08-25
local environment to override included .env

ndeloof:
as we include another compose.yaml file and parse implicit or explicit sibling `.env` file, upper-level project environment should override values (just like `FOO=BAR docker compose up` does)

see https://github.com/docker/compose/issues/10913


milas (approved):
LGTM! Want to update `TestLoadsWithInclude` to cover this case? 😉

PR #451, opened 2023-08-29
add support for multi-document yaml

ndeloof:
Allows a compose.yaml file to define a multi-document yaml content. Documents all get merged into a single compose model.
I don't expect this to be used by users directly, but to allow distribution of a set of compose files without the need to manage file paths, just as a single yaml stream with all documents inlined in expected order.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -17,8 +17,10 @@
 package loader
 
 import (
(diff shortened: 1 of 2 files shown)


milas (approved):
Nice!

PR #459, opened 2023-09-19
loader: report project structure

milas:
With `include`'d Compose files, it's important to know the origin of each service to be able to properly resolve paths to them, for example.

A new `FileMeta` field on the `Project` type provides detailed access to the underlying file structure/hierarchy.

This replaces `IncludeReferences` and `ComposeFiles`, which were inconsistent and not comprehensive.

## What Problem Does This Solve
In Compose itself, when we use `x-` fields for non-stable features, the "load" logic lives in the `docker/compose` codebase until stabilized.

That means it needs to resolve in relative paths there -- `compose-go` can't help because it doesn't know about the fields to begin with. However, if they came from an included file, `IncludeReferences` has the `ProjectDirectory` BUT there's no way to know where a service came from! The `FileMeta` tree combines `ComposeFiles` + `IncludeReferences` and, critically, additionally populates service back-references.

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -408,7 +408,6 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 		return nil, err
 	}
 
-	project.ComposeFiles = configPaths
 	return project, nil
 }
 
--- a/cli/options_test.go
+++ b/cli/options_test.go
@@ -223,7 +223,7 @@ func TestProjectComposefilesFromSetOfFiles(t *testing.T) {
 	p, err := ProjectFromOptions(opts)
 	assert.NilError(t, err)
 	absPath, _ := filepath.Abs(filepath.Join("testdata", "simple", "compose.yaml"))
-	assert.DeepEqual(t, p.ComposeFiles, []string{absPath})
+	assert.DeepEqual(t, p.FileMeta.Path, absPath)
 }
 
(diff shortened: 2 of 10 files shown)


milas:
Open to suggestions here: I actually originally had this as an entirely separate object returned along _with_ the project, but ergonomically that's impractical to propagate deep down into Compose itself.

Hopefully this is a reasonable compromise. I'd love to revamp all this soon 🙃

ndeloof:
LGTM in terms of codebase

> With include'd Compose files, it's important to know the origin of each service to be able to properly resolve paths to them, for example.

About this assumption: included service _should_ have all path resolved already when resource is copied into project, so why is this needed?

glours:
@ndeloof IIUC, in the case of service managed by the `watch` mode and coming from a `include`d file, we lose the original directory from where the relative paths defined in `watch` section should be applied

ndeloof:
ok, so basically because we miss https://github.com/compose-spec/compose-go/pull/461/files#diff-3e93eb737303c2fd36bba9ddd3b891ad9fb906fda7dbf9e377ceacb5b89bd943R118-R124 by using a custom extension, make more sense now

milas:
(Marking this as draft temporarily so it doesn't get merged - @ndeloof's PR @ #461 resolves the `watch` + `include` issue. I think this might be useful for the future, but not critical for today.)

PR #461, opened 2023-09-20
introduce `develop` section

ndeloof:
This introduces support for `develop` section as defined by https://github.com/compose-spec/compose-spec/pull/253

### DIFF
--- a/loader/paths.go
+++ b/loader/paths.go
@@ -115,6 +115,13 @@ func ResolveServiceRelativePaths(workingDir string, s *types.ServiceConfig) {
 		}
 		s.Volumes[i].Source = resolveMaybeUnixPath(workingDir, vol.Source)
 	}
+
+	if s.Develop != nil {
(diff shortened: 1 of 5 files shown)


milas (approved):

loader/paths.go
```diff
@@ -115,6 +115,13 @@ func ResolveServiceRelativePaths(workingDir string, s *types.ServiceConfig) {
 		}
 		s.Volumes[i].Source = resolveMaybeUnixPath(workingDir, vol.Source)
 	}
+
+	if s.Develop != nil {
+		for i, w := range s.Develop.Watch {
+			w.Path = absPath(workingDir, w.Path)
```
milas, line 121: Noting that this should fix the `include` + `watch` issue since we're now resolving the path at the right place.

My PR in #459 can still be useful for the future if/when we have other cases where we need to resolve paths outside of a load, but isn't critical anymore

PR #462, opened 2023-09-21
dotenv: fix an unintended buffer copy

nicks:
before this change: the test allocates ~10MB
after this change: the test allocates ~300KB

fixes https://github.com/docker/compose/issues/10951


ndeloof (approved):
💯

ndeloof:
nice catch, thanks for digging into this issue

glours (approved):

PR #464, opened 2023-09-22
merge yaml files as plain map[string]interface trees

ndeloof:
Implement the Compose file merge (aka "override") logic on a plain `map[string]interface{}` trees so that we can process compose files before those are unmarshalled into go structs


ndeloof:
replaced by https://github.com/compose-spec/compose-go/pull/468

PR #465, opened 2023-09-27
introduce DecodeMapstructure to allow type to define custom decode logic

ndeloof:
This introduces a reflexion-based mechanism, comparable to yaml/json `Unmarshaler` interface, for types to define their own logic to convert value into internal data. This allows to have a single implementation of data type normalization, vs one declared as a mapstructure hook + one declared as a interpolation type cast (even if those redirect to a central func).

This also will make it easier to convert those into plain `yaml.Unmarshaler` if/once we get the parse+merge logic implemented _before_ converting to go structs

note: only implemented in a few structs, a few more are eligible to this approach


glours (approved):
👌

PR #467, opened 2023-09-28
define Options type for (driver) options we pass as-is to runtime

ndeloof:
introduce `Options` as an explicit type for `map[string]string` driver options we pass as-is to container runtime

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -535,6 +535,9 @@ func (m Mapping) Merge(o Mapping) Mapping {
 	return m
 }
 
+// Options is a mapping type for options we pass as-is to container runtime
+type Options map[string]string
+
 // Labels is a mapping type for labels
 type Labels map[string]string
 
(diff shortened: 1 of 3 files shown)


types/types.go
```diff
@@ -979,18 +982,18 @@ type IPAMPool struct {
 	Subnet             string                 `yaml:"subnet,omitempty" json:"subnet,omitempty"`
 	Gateway            string                 `yaml:"gateway,omitempty" json:"gateway,omitempty"`
 	IPRange            string                 `yaml:"ip_range,omitempty" json:"ip_range,omitempty"`
-	AuxiliaryAddresses map[string]string      `yaml:"aux_addresses,omitempty" json:"aux_addresses,omitempty"`
```
ndeloof, line 982: I wonder this one should be defined as a `map[string]net.IPAddr` and we rely on `ParseIP` for parsing and validation

glours (approved):
LGTM

PR #468, opened 2023-10-09
Load pipeline on yaml tree vs go structs

ndeloof:
This replaces use of mergo and various reflection hacks with a plain yaml tree (`map[string]any`) processing


ndeloof:
pushed to https://github.com/compose-spec/compose-go/tree/v2

Issue #471, opened 2023-10-13
loader panic


nicks:
Add the following test to loader_test:

```
func TestBadServiceConfig(t *testing.T) {
	yaml := `name: scratch
services:
  redis:
    image: redis:6.2.6-alpine
    network_mode: bridge
    networks:
      gratheon: null
networks:
  gratheon:
    name: scratch_gratheon
`
	_, _ = LoadWithContext(context.Background(), types.ConfigDetails{
		ConfigFiles: []types.ConfigFile{
			{
				Content: []byte(yaml),
			},
		},
	})
}
```

Expected result:
Loader emits a reasonable error

Actual result:
Loader panics

```
--- FAIL: TestBadServiceConfig (0.00s)
panic: assignment to entry in nil map [recovered]
	panic: assignment to entry in nil map
```

nicks:
distilled from https://github.com/tilt-dev/tilt/issues/6249

PR #472, opened 2023-10-13
loader: fix a panic

nicks:
fixes https://github.com/compose-spec/compose-go/issues/471


nicks:
(originally was going to only file the issue, but it turned out to be easy enough to fix)

ndeloof (approved):
LGTM
I really miss constructors in go so that we can have default values and empty maps set vs nil

PR #473, opened 2023-10-26
introduce compose-spec command line for compose file validation

ndeloof:
As part of compose-go v2, we consider introducing a command line to manage compose files and validate those, somehow comparable to `docker compose config` command but not tied to Docker implementation.


glours (approved):
LGTM

PR #481, opened 2023-11-03
make scale a *int as default is 1 (not 0)

ndeloof:
as default is not 0, `omitempty` doesn't apply and this require code to manage this default value.
using an `*int` we can distinguish unset value (scale=1) vs service explicitly set to scale=0 (many users rely on this to disable service)


types/types.go
```diff
@@ -147,7 +146,6 @@ func (s ServiceConfig) MarshalYAML() (interface{}, error) {
 func (s ServiceConfig) MarshalJSON() ([]byte, error) {
 	type t ServiceConfig
 	value := t(s)
-	value.Scale = 0 // deprecated, but default value "1" doesn't match omitempty
 	return json.Marshal(value)
 }
```
milas, line 150: Can get rid of this now, it's not doing anything custom anymore

milas (approved):

PR #484, opened 2023-11-07
add support for inline config

ndeloof:
This adds support for inlined config (https://github.com/compose-spec/compose-spec/pull/429)


glours (approved):
LGTM

PR #486, opened 2023-11-08
add support for build.ulimits

ndeloof:
implement https://github.com/compose-spec/compose-spec/pull/433


schema/compose-spec.json
```diff
@@ -872,4 +872,4 @@
       }
     }
   }
-}
+}
```
glours, line 875: nit: new line

glours (approved):
Just a small nit, no need to fix it

PR #490, opened 2023-11-16
validate watch.path is set to a valid path

ndeloof:
this bump compose-spec schema (watch and path are required) + adds a validation step to check the configured path exists

closes https://github.com/docker/compose/issues/11182


glours (approved):
LGTM

PR #493, opened 2023-11-20
don't check path is an actual file

ndeloof:
after https://github.com/compose-spec/compose-go/pull/490 I changed my mind, considering this use-case:
as a team has a compose.yaml file, developers have `develop` section set, but ops only rely on published images, the latter won't be able to run `compose up` as they will get watch.path validation error, which isn't relevant for their usage.
So, better check valid path when actually used by some command.


glours (approved):
Sounds good to me

PR #495, opened 2023-11-21
add a temporary job to sync v1 branch compose-spec.json with the spec repo one

glours:
We lost the PR creation for branch v1 when `compose-spec.json` diverge from the Compose spec

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


ndeloof:
as discussed with the team, would be better in practice we sync the other way, so that changes in compose-go includes both schema update + code then we update the compose-spec repo once approved

glours:
Close, replace by https://github.com/compose-spec/compose-spec/pull/455

PR #496, opened 2023-11-21
update compose-spec.json with path and action as required field for develop

glours:
Need to merge this one to quickly unblock #494


glours:
Validation error is part of the #494

PR #499, opened 2023-11-22
Permit '=' separator and '[ipv6]' in 'extra_hosts'.

robmry:
Fixes docker/cli#4648

Align the format of `extra_hosts` strings with updated `--add-hosts` options in the docker CLI and buildx - by permitting `host=ip` in addition to `host:ip`, and allowing square brackets around the address.

For example:

```
    extra_hosts:
      - "my-host1:127.0.0.1"
      - "my-host2:::1"
      - "my-host3=::1"
      - "my-host4=[::1]"
```

Added `HostsList` unit tests for the various input formats.

### DIFF
--- a/types/hostList.go
+++ b/types/hostList.go
@@ -20,28 +20,33 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 )
 
 // HostsList is a list of colon-separated host-ip mappings
 type HostsList map[string]string
 
-// AsList return host-ip mappings as a list of colon-separated strings
-func (h HostsList) AsList() []string {
+// AsList returns host-ip mappings as a list of strings, using the given
+// separator. The Docker Engine API expects ':' separators, the original format
+// for '--add-hosts'. But an '=' separator is used in YAML/JSON renderings to
+// make IPv6 addresses more readable (for example "my-host=::1" instead of
+// "my-host:::1").
+func (h HostsList) AsList(sep string) []string {
 	l := make([]string, 0, len(h))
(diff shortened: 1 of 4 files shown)


types/hostList.go
```diff
@@ -58,9 +59,21 @@ func (h *HostsList) DecodeMapstructure(value interface{}) error {
 		}
 		*h = list
 	case []interface{}:
-		*h = decodeMapping(v, ":")
+		*h = decodeMappingSeps(v, []string{"=", ":"})
```
ndeloof, line 62: could change `decodeMapping` to accept a variatic `sep` parameter so we don't need another implementation

ndeloof (approved):
LGTM
a minor note on extending `decodeMapping` to support both scenarios

types/hostList.go
robmry: Ah, thank you - that's a better idea! Will do ...

ndeloof:
to promote use of `=` , `HostsList.AsList()` should also use it when rendering canonical compose.yaml with `docker compose config`

robmry:
> to promote use of `=` , `HostsList.AsList()` should also use it when rendering canonical compose.yaml with `docker compose config`

I'm still figuring out how all this stuff fits together - but the engine API won't understand `=`, it's up to the CLI/buildx/compose clients to convert back to `:` (less human-readable, but a stable API that avoids problems with an old engine vs. new client).

So, I think, if the `docker compose config` is what ends up being sent to the API (?), it'll need to stick with the `=`?

PR #506, opened 2023-12-07
introduce general-purpose service dependency graph traversal functions

ndeloof:
This introduces in compose-go the `InDependencyOrder` utility method which was designed for docker/compose
Comes with a revisited (simplified) implementation, + a new `CollectInDependencyOrder` utility to run function on services but also collect results into a `map[string]T`

Also revisit `WithServices` (same purpose, but not running in dependency order neither concurrently) to pass service name to function, and also adopt functional options pattern to declare dependency policy - could be used in future to introduce more options

### DIFF
--- a/utils/collectionutils.go
+++ b/utils/collectionutils.go
@@ -18,14 +18,12 @@ package utils
 
 import (
 	"golang.org/x/exp/constraints"
+	"golang.org/x/exp/maps"
 	"golang.org/x/exp/slices"
 )
 
 func MapKeys[T constraints.Ordered, U any](theMap map[T]U) []T {
-	result := make([]T, 0, len(theMap))
-	for key := range theMap {
-		result = append(result, key)
-	}
+	result := maps.Keys(theMap)
 	slices.Sort(result)
 	return result
 }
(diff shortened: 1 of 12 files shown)


utils/set.go
glours, line 1: And why not an immutable Set? 😄

utils/collectionutils.go
```diff
@@ -16,13 +16,17 @@
...
+)
 
-func MapKeys[T comparable, U any](theMap map[T]U) []T {
-	var result []T
+func MapKeys[T constraints.Ordered, U any](theMap map[T]U) []T {
+	result := make([]T, 0, len(theMap))
 	for key := range theMap {
 		result = append(result, key)
 	}
+	slices.Sort(result)
```
glours, line 29: Now we can use https://pkg.go.dev/golang.org/x/exp/maps#Keys
If we want to keep it sorted: 
```suggestion
func MapKeys[T constraints.Ordered, U any](theMap map[T]U) []T {
	result := slices.Sort(maps.Keys(theMap))
```

glours (approved):
LGTM

utils/set.go
ndeloof: this is the code we had in docker/compose, but I'm fine to change it to another implementation

PR #513, opened 2023-12-20
fix #512 revert err check

### LINKED ISSUE #512: Regression: Decode errors are ignored
Regression introduced in 7e391a6 @ndeloof 

https://github.com/compose-spec/compose-go/blob/7e391a6500eeb366211891b2f6215412058de9a0/loader/loader.go#L342

This piece of code got deleted:

https://github.com/compose-spec/compose-go/commit/7e391a6500eeb366211891b2f6215412058de9a0#diff-8e9af1e483632eac17610237caaa5f879ea7bfb3e1e1ea228e45a17c770b9432L303-L305

lionello:
Restore the `if err` that was likely unintentionally deleted in https://github.com/compose-spec/compose-go/commit/7e391a6500eeb366211891b2f6215412058de9a0

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -342,6 +342,9 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 				if err == io.EOF {
 					break
 				}
+				if err != nil {
+					return nil, err
+				}
 				if err := processRawYaml(raw, processor); err != nil {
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

ndeloof:
thanks so much for catching this mistake of mine and offer a quick fix !

PR #515, opened 2023-12-20
Update schema json

lionello:
Should fix CI failures in #513


ndeloof:
this is actually already covered by https://github.com/compose-spec/compose-go/pull/509

PR #516, opened 2023-12-20
Read config file content if missing

### LINKED ISSUE #489: Project name not checked when using ToConfigFiles
`ToConfigFiles` does not read the content of the Compose files. In the `loader.projectName` function when checking for the existence of the top-level "name" key, the `Content` field is not set, so the read fails silently and never actually checks the "name" value.

https://github.com/compose-spec/compose-go/blob/main/loader/loader.go#L482

A workaround is to explicitly set the `Content` field instead of using `ToConfigFiles`.

aksiksi:
This ensures that project name validation occurs in cases where the content is missing (e.g., when constructed via `ToConfigFiles`).

I opted to do the read during project name parsing for two reasons:

1. It addresses all cases where `Content` is not set (not just `ToConfigFiles`).
2. It does not require a (breaking?) change to the `ToConfigFiles` function.

I also had to fix a few tests that were not setting the project correctly. 

Fixes #489.

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -978,7 +978,7 @@ services:
 	})
 
 	assert.DeepEqual(t, actual.Services["db"].Build, &types.BuildConfig{
-		Context:    wd + "/db",
+		Context:    filepath.Join(wd, "db"),
 		Dockerfile: "Dockerfile",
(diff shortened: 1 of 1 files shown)


ndeloof (approved):

ndeloof:
please sign-off your commit (required by legals)

PR #518, opened 2023-12-27
Make public functions of Project type immutable

glours:
⚠️ Breaking changes ⚠️ 

**What I did**

- Add `github.com/mitchellh/copystructure` to make deep copy of `Project` struct
- Make all `Project` public functions which mutate `Project` state immutabl, returning a new instance of `Project` and usage a common `With` prefix for each of them
- Replace the `Project.WithProject` function by a `Project.ForEachProject` function
- Remove the unused `types.set` type

**Related issue**
https://docker.atlassian.net/browse/COMP-380


types/project.go
```diff
@@ -572,7 +618,15 @@ func (p Project) ResolveServicesEnvironment(discardEnvFiles bool) error {
...
 			service.EnvFiles = nil
 		}
-		p.Services[i] = service
+		newProject.Services[i] = service
 	}
-	return nil
+	return newProject, nil
+}
+
+func (p *Project) projectDeepCopy() (*Project, error) {
```
ndeloof, line 626: why not just `deepCopy`?

glours: useless gif

ndeloof (approved):

types/project.go
```diff
@@ -185,13 +188,19 @@ func (p *Project) AllServices() Services {
 
 type ServiceFunc func(name string, service ServiceConfig) error
 
-// WithServices run ServiceFunc on each service and dependencies according to DependencyPolicy
-func (p *Project) WithServices(names []string, fn ServiceFunc, options ...DependencyOption) error {
+// WithServices runs ServiceFunc on each service and dependencies according to DependencyPolicy
+// It returns a new Project instance with the changes and keep the original Project unchanged
+func (p *Project) WithServices(names []string, fn ServiceFunc, options ...DependencyOption) (*Project, error) {
```
milas, line 193: This method is the only one I think is odd - the returned project option is always identical to the original.

I think it might make sense to have a pair of functions:
 * `ForEach(func(name string, svc ServiceConfig) error) error` - functional loop helper, does not copy project
 * `MutateServices(func(name string, svc *ServiceConfig) error) (*Project, error)` - clone project, allowing caller to get a mutable reference to each service in the process

types/project.go
```diff
@@ -437,32 +464,46 @@ func (p *Project) ForServices(names []string, options ...DependencyOption) error
...
-	p.Services = enabled
-	return nil
+	newProject.Services = enabled
+	return newProject, nil
 }
 
-func (p *Project) DisableService(service ServiceConfig) {
+// DisableService removes from the project model the given service and its references in all dependencies
+// It returns a new Project instance with the changes and keep the original Project unchanged
+func (p *Project) DisableService(service ServiceConfig) (*Project, error) {
```
milas, line 478: minor: it's kind of weird that this one takes a `ServiceConfig` instead of a name like all the others (but that's not new)

types/project.go
```diff
@@ -572,7 +618,15 @@ func (p Project) ResolveServicesEnvironment(discardEnvFiles bool) error {
...
 	}
-	return nil
+	return newProject, nil
+}
+
+func (p *Project) deepCopy() (*Project, error) {
+	instance, err := copystructure.Copy(p)
+	if err != nil {
+		return nil, err
+	}
```
milas, line 630: In theory, this should not fail, right? It might make sense for this to `panic()` to avoid carrying the `error` everywhere.

types/project.go
```diff
@@ -407,25 +433,26 @@ func IgnoreDependencies(options *withServicesOptions) {
 	options.dependencyPolicy = ignoreDependencies
 }
 
-// ForServices restrict the project model to selected services and dependencies
-func (p *Project) ForServices(names []string, options ...DependencyOption) error {
+// ForServices restricts the project model to selected services and dependencies
+// It returns a new Project instance with the changes and keep the original Project unchanged
+func (p *Project) ForServices(names []string, options ...DependencyOption) (*Project, error) {
```
milas, line 438: (see other comment re: getting rid of `error` from `deepCopy()`)

I am wondering if it makes sense to ignore unknown names here, so that this method can return `*Project`, which would make it much more ergonomic to use: `p = p.ForServices(svc)`.

In particular, this could allow chaining, e.g.
```go
p.ForServices(svcs).ForEach(...)
```

milas (approved):
LGTM, some comments but nothing majorly concerning.

Two other high-level thoughts:

1. `mitchellh` is no longer doing Go dev - he recently archived a bunch of libs; I'm not super concerned since deep copying is an internal detail and doesn't pollute the API, more of an FYI

2. The naming & arguments of all these functions is pretty inconsistent, since this is all breaking API changes anyway, it might make sense to do a pass for consistency, e.g. make them all `WithFoo()` style.

ndeloof:
> mitchellh is no longer doing Go dev

.. which is sad as we rely on mapstructure for parsing. Maybe we should revisit this to only rely on yaml.Unmarshall.
 We could implement `deepClone` without such a reflexion framework, by plain hand-written clone code

milas:
Yeah...at least `mapstructure` does seem to be maintained with recent commits and is very popular. In either case, I'm not super worried – we can also always fork them (or move to a community fork) if the need arises

types/project.go
glours: Or we can introduce an `Either` type 😇 , WDYT?

ndeloof: > I am wondering if it makes sense to ignore unknown names here

this would be a bug then if we silently ignore incorrect parameter being set

types/project.go
ndeloof: Agree. As we now use `WithXX` fr mutating functions, this name is not relevant anymore. `ForEach` would be a better one. We then can have a new `WithServiceFn(func(ServiceConfig) ServiceConfig)` for service mutation, if relevant (afaik we don't use this in docker/compose)

ndeloof (approved):

ndeloof (approved):

types/project.go
```diff
@@ -186,21 +186,16 @@ func (p *Project) AllServices() Services {
...
 
-// WithServices runs ServiceFunc on each service and dependencies according to DependencyPolicy
-// It returns a new Project instance with the changes and keep the original Project unchanged
-func (p *Project) WithServices(names []string, fn ServiceFunc, options ...DependencyOption) (*Project, error) {
-	newProject, err := p.deepCopy()
-	if err != nil {
-		return nil, err
-	}
+// ForEachService runs ServiceFunc on each service and dependencies according to DependencyPolicy
+// The fn ServiceFunc should not mutate its ServiceConfig attribute
```
milas, line 192:
```suggestion
// ForEachService runs ServiceFunc on each service and dependencies according to DependencyPolicy.
//
// No copies are created; any mutation of `*ServiceConfig` by the fn `ServiceFunc` will apply to this `*Project` object.
```

My 2c: either we should pass `ServiceConfig::deepCopy()` to `ServiceFunc` to make it impossible for callers to mutate the project, or accept that this function allows for in-place mutation and encourage callers to call `proj.Clone()` themselves if needed.

milas (approved):
lgtm!

types/project.go
ndeloof: we can indeed clone the ServiceConfig as a foolguard, anyway any attempt we make in this direction can only be a recommendation, as attributes are exported so third party can do mutation in many places (so does docker/compose, we don't use this function so much, and have tons of `for s := range project.Services`)

milas:
I fixed the DCO status and set the option in the repo settings so it doesn't happen again with accepting PR comments:

Screenshot 2024-01-10 at 11 20 11 AM

types/project.go
glours: @milas done, I added a `ServiceConfig.deepCopy` function and use it when calling the `fn` ServiceFunc 👍

glours:
> I fixed the DCO status and set the option in the repo settings so it doesn't happen again with accepting PR comments:
> 
> Screenshot 2024-01-10 at 11 20 11 AM

Arff I just push force without pulling your changes my bad 🤦

ndeloof (approved):

PR #522, opened 2024-01-12
neutralize path separator when used in service name

ndeloof:
`tree.Path` must distinguish parts in a path vs names using the path separator char.
This quick fix replaces path separator with another char I don't expect anyone would use for resource name

closes https://github.com/docker/compose/issues/11336


glours (approved):
😂

ndeloof:
a nicer approach is to redefine `tree.Path` as a slice of parts, so that parts can be any string. I'm looking into this, but can't guarantee a PR to be ready short-terms

erayaydin:
:ghost: funny touch

PR #524, opened 2024-01-15
fix regression in cycle detection

ndeloof:
cycle tracker must fork as we walk through `extends` services, to be considered as a tree 

closes https://github.com/docker/compose/issues/11343


glours (approved):
Sounds good to me

PR #526, opened 2024-01-15
fix regression merging overrides

ndeloof:
`service.networks` can be defined as a sequence or mapping. Sequence syntax is equivalent to a mapping with value=null. Merge must operate on the mapping representation

includes test case

closes https://github.com/docker/compose/issues/11339

### DIFF
--- a/loader/override_test.go
+++ b/loader/override_test.go
@@ -1,0 +1,63 @@
+/*
+   Copyright 2020 The Compose Specification Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


loader/loader.go
```diff
@@ -327,6 +327,11 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 				}
 			}
 
+			cfg, err = transform.Canonical(cfg)
```
jhrotko, line 330: should we add tests for this case?

laurazard (approved):
LGTM ✅ (some tests for this would be nice 👍)

glours (approved):
LGTM

PR #527, opened 2024-01-15
Fix regression with additional_context short syntax

ndeloof:
need to convert additional_context to canonical (mapping) syntax before relative paths get resolved

closes https://github.com/docker/compose/issues/11351


jhrotko (approved):
LGTM

Issue #519, opened 2023-12-27
Build error on override/merge.go line 129


lionello:
https://github.com/compose-spec/compose-go/blob/8585af8ec50b47489c1e7304dde3a7577aa658e4/override/merge.go#L129

Latest version of golang/x/exp changes type of `SortFunc`: it now returns an `int`. See https://cs.opensource.google/go/x/exp/+/302865e7556b4ae5de27248ce625d443ef4ad3ed:slices/sort.go;dlc=d98519c114953aa372c373aa4c7122a39d7932af

When `golang.org/x/exp` is too new, you get this build error:
```
../../../go/pkg/mod/github.com/compose-spec/compose-go/v2@v2.0.0-beta.2/override/merge.go:129:24: type func(a any, b any) bool of func(a, b any) bool {…} does not match inferred type func(a any, b any) int for func(a E, b E) int
```

I fixed it by downgrading the exp package:
```
go get golang.org/x/exp@613f0c0eb8a1
```

PR #531, opened 2024-01-17
Bump golang minimal version to 1.21 and 1.22

jhrotko:
In order to update the version of https://pkg.go.dev/golang.org/x/exp@v0.0.0-20240112132812-db7319d0e0e3/slices we need to update the `slices.SortFunc` to be compatible withy the latest version.

Related with https://github.com/compose-spec/compose-go/issues/519


override/merge.go
```diff
@@ -179,7 +180,7 @@ func convertIntoSequence(value any) []any {
 			i++
 		}
 		slices.SortFunc(seq, func(a, b any) int {
-			return compare(a.(string), b.(string))
```
ndeloof, line 182: please also remove this func, as we won't need the conditional build anymore

jhrotko, line 182: So, I am not sure, should we use `compare` or `cmp.Compare`? @ndeloof

ndeloof: `cmp.Compare` should be safe (?), at least we already use it on go1.21+, so better just remove the compare/compare120 hack

ndeloof (approved):

glours (approved):

PR #535, opened 2024-01-19
Fix ulimit parsing

ndeloof:
canonical transformation must keep `ulimit` as an int or struct to match json schema (validated after a merge)
converting `int` to internal `single` field must take place during the decode phase, as this is specific to go-structs design

fixes https://github.com/docker/compose/issues/11353#issuecomment-1899835614


glours (approved):
LGTM

PR #538, opened 2024-01-19
always resolve paths relative to extending file's working dir

ndeloof:
closes https://github.com/docker/compose/issues/11377
closes https://github.com/docker/compose/issues/11379

note: custom resource loaders defined by docker/compose will have to implement `Dir(path string)` returning (absolute) path for file within the remote download cache

### DIFF
--- a/loader/extends.go
+++ b/loader/extends.go
@@ -26,7 +26,7 @@ import (
 	"github.com/compose-spec/compose-go/v2/types"
 )
 
-func ApplyExtends(ctx context.Context, dict map[string]any, workingdir string, opts *Options, tracker *cycleTracker, post ...PostProcessor) error {
+func ApplyExtends(ctx context.Context, dict map[string]any, opts *Options, tracker *cycleTracker, post ...PostProcessor) error {
 	a, ok := dict["services"]
 	if !ok {
 		return nil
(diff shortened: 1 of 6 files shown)


loader/extends.go
```diff
@@ -71,14 +71,15 @@ func ApplyExtends(ctx context.Context, dict map[string]any, workingdir string, o
...
+				localdir := filepath.Dir(local)
+				relworkingdir := loader.Dir(path)
+
 				extendsOpts := opts.clone()
+				extendsOpts.ResourceLoaders = append([]ResourceLoader{}, opts.ResourceLoaders...)
+				// replace localResourceLoader with a new flavour, using extended file base path
+				extendsOpts.ResourceLoaders[len(opts.ResourceLoaders)-1] = localResourceLoader{
+					WorkingDir: localdir,
+				}
 				extendsOpts.ResolvePaths = true
```
glours, line 83: This line is the root cause of the https://github.com/docker/compose/issues/11377
It was introduced to remove potential import cycles https://github.com/compose-spec/compose-go/commit/714c186dd139b0e732a9441c4445a6836a8d10d4

glours (commented):
Can we keep the value of `opts.ResolvePaths` for the `extendsOpts`?
Its value can be set to false only from `config` command on Compose

loader/extends.go
ndeloof: `ResolvePaths` MUST be set so that `./env` in a compose file that is used by `extends: file=./foo/bar/compose.yaml` is resolved as `./foo/bar/env` and can be used as expected by the caller

Used by extend, PathResolution resolves _relative_ paths from working directory, which is itself relative

glours (approved):
LGTM, I wasn't testing with the fixed version 🤦

PR #545, opened 2024-01-23
add go binding for storage_opt

ndeloof:
`storage_opt` is already defined by compose-spec schema but we didn't had binding set in go structs for it

see https://github.com/docker/compose/issues/11395

### DIFF
--- a/loader/full-example.yml
+++ b/loader/full-example.yml
@@ -273,7 +273,8 @@ services:
     stop_grace_period: 20s
 
     stop_signal: SIGUSR1
-
+    storage_opt:
+      size: "20G"
     sysctls:
       net.core.somaxconn: 1024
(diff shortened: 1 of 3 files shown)


jhrotko (approved):
LGTM! I was checking our docs page for [v3](https://docs.docker.com/compose/compose-file/compose-file-v3/) and there is no reference for `storage_opt`. Only in [v2](https://docs.docker.com/compose/compose-file/compose-file-v2/) docs. Is this something that needs to be added?

ndeloof:
@jhrotko compose file v2 vs v3 is obsolete, compose-specification merged those into a single (version-less) file format with all legacy attributes supported

glours (approved):
LGTM

PR #547, opened 2024-01-24
process `extends` recursively

ndeloof:
fixes https://github.com/docker/compose/issues/11394

issue takes place when there's a chain of extending services:
```yaml
services:
  level1:
    build:
      context: ..

  level2:
    extends: level1

  service:
    extends: level2
    build:
      target: target
```

The root cause for this issue is that we try to extends a service before the compose model has been fully resolved. Typically loading this illustration example, if first service (which go map iteration makes random) to be processed is `service`, it will merge with sibling `level2`. This results in a `build` mapping defined without a `context`, and will default to `.`. But if `level2` is loaded first, it will have `build.context` set to `..`, and then `service` will do as well.

This PR changes the logic so that we force `extends` to be applied recursively on services using new func `applyServiceExtends`. This one will process a service to apply `extends` if needed (otherwise is no-op). Once it has identified the base service used by extends, it executes recursively to enforce this one to be fully defined

### DIFF
--- a/loader/extends.go
+++ b/loader/extends.go
@@ -48,6 +48,9 @@ func ApplyExtends(ctx context.Context, dict map[string]any, opts *Options, track
 
 func applyServiceExtends(ctx context.Context, name string, services map[string]any, opts *Options, tracker *cycleTracker, post ...PostProcessor) (any, error) {
 	s := services[name]
+	if s == nil {
+		return nil, nil
+	}
 	service, ok := s.(map[string]any)
 	if !ok {
 		return nil, fmt.Errorf("services.%s must be a mapping", name)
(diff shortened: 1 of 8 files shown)


loader/extends.go
```diff
@@ -35,96 +35,120 @@ func ApplyExtends(ctx context.Context, dict map[string]any, opts *Options, track
...
+		}
+		localdir := filepath.Dir(local)
+		relworkingdir := loader.Dir(path)
+
+		extendsOpts := opts.clone()
+		extendsOpts.ResourceLoaders = append([]ResourceLoader{}, opts.ResourceLoaders...)
+		// replace localResourceLoader with a new flavour, using extended file base path
+		extendsOpts.ResourceLoaders[len(opts.ResourceLoaders)-1] = localResourceLoader{
+			WorkingDir: localdir,
 		}
```
milas, line 128: Instead of assuming that `localResourceLoader` is the last in the slice, it might be better to loop through and attempt to cast them until you find it and then replace it

milas (approved):
👍🏻

loader/extends.go
ndeloof: I introduced `RemoteResourceLoaders` to check this implicit rule

PR #548, opened 2024-01-24
fix: handle more complex fields during merge

glours:
Improve the overall merge process for specific attribute format and uniqueness for lists and sequences

closes https://github.com/docker/compose/issues/11417


milas (changes requested):
LGTM but note that we need the same for `services.*.annotations`, `services.*.extra_hosts`, and `services.*.sysctls` as well 😬 

There might be others, but those are all k-v list or maps, so were easy to identify.

ndeloof (approved):

logopk:
I have noticed problems with duplicate IP subnets. docker-compose version 2.24.2 and 2.24.3

```
  default:
    driver: bridge
    ipam:
     config:
       - subnet: 172.18.0.0/29  
```
override:
```
  default:
    driver: bridge
    ipam:
     config:
       - subnet: 172.23.0.0/29
```

docker-compose config shows

```
  default:
    name: apache_default
    driver: bridge
    ipam:
      config:
        - subnet: 172.18.0.0/29
        - subnet: 172.23.0.0/29
```

Error message on up is:

```
 ✘ Network apache_default  Error                                                                                                                                  0.0s 
failed to create network apache_default: Error response from daemon: bridge driver doesn't support multiple subnets
```

This was not a problem so far. config shows the overridden subnet and up starts the container in this subnet

```
  default:
    name: apache_default
    driver: bridge
    ipam:
      config:
      - subnet: 172.23.0.0/29
```

if it would be the same subnet in both files it would not be merged:
```
  default:
    name: apache_default
    driver: bridge
    ipam:
      config:
        - subnet: 172.22.0.0/29
        - subnet: 172.22.0.0/29
```

leading to a more bizarre error:

```
 ✘ Network apache_default  Error                                                                                                                                  0.0s 
failed to create network apache_default: Error response from daemon: Pool overlaps with other one on this address space
```

override/merge.go
```diff
@@ -180,6 +195,16 @@ func mergeUlimit(_ any, o any, p tree.Path) (any, error) {
 	return o, nil
 }
 
+func mergeIPAMConfig(c any, o any, path tree.Path) (any, error) {
+	right := convertIntoMapping(c.([]any)[0], nil)
```
ndeloof, line 199: I don't get the logic here. Why only merge the first element ?
IIUC we should merge IPAM configs within the same network segment (same `subnet`) and append the others

glours: Rohh my bad 🤦

override/merge_networks_test.go
```diff
@@ -0,0 +1,214 @@
...
+func Test_mergeYamlNetworks(t *testing.T) {
+	assertMergeYaml(t, `
+services:
+  test:
+    image: foo
+networks:
+  network1:
+    ipam:
+      config:
+        - subnet: 172.28.0.0/16
```
ndeloof, line 150: tes should include another subnet which should be kept unchanged, + another that should be appended, so we cover all use-cases

override/merge.go
glours: I'll fix that Monday 😞

milas (commented):
LGTM besides the already-commented on IPAM merge behavior

PR #549, opened 2024-01-24
Fix #520 project name not available for lookup

### LINKED ISSUE #520: Environment copied before being mutated
This line initializes the `Environment`, when not initialized by the caller:
https://github.com/compose-spec/compose-go/blob/8585af8ec50b47489c1e7304dde3a7577aa658e4/loader/loader.go#L269

But the environment gets copied (implicitly) earlier on this line:
https://github.com/compose-spec/compose-go/blob/8585af8ec50b47489c1e7304dde3a7577aa658e4/loader/loader.go#L249

So, when `Environment` is `nil`, the set of `COMPOSE_PROJECT_NAME` in the map is not visible by the call to `LookupValue`.

A potential fix is to change `func (cd ConfigDetails) LookupEnv(…)` to a pointer receiver to avoid the copy.

lionello:
This PR fixes #520

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -2969,3 +2969,73 @@ services:
 	})
 	assert.NilError(t, err)
 }
+
+func TestLoadProjectName(t *testing.T) {
+	const projectName = "testproject"
+
+	tests := []struct {
+		name    string
+		env     map[string]string
+		options func(*Options)
+		wantErr string
+	}{
+		{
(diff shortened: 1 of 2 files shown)


lionello:
@ndeloof check the failing test to see what I meant in #520

ndeloof (approved):

ndeloof:
got it, thanks

PR #552, opened 2024-01-26
!override can be used to replace value

ndeloof:
Introduce `!override` so one can replace any entry in compose.override.yaml, while `!reset` would remove but not allow to re-define
i.e:
- `foo: !reset null` removes `foo` attribute from merged yaml (value is ignored)
- `foo: !override hello` overrides any previous `foo` attribute value with new value `"hello"`

**TODO**: document this on the compose spec

fixes https://github.com/docker/compose/issues/11408

### DIFF
--- a/loader/reset.go
+++ b/loader/reset.go
@@ -44,6 +44,10 @@ func (p *ResetProcessor) resolveReset(node *yaml.Node, path tree.Path) (*yaml.No
 		p.paths = append(p.paths, path)
 		return nil, nil
 	}
+	if node.Tag == "!override" {
+		p.paths = append(p.paths, path)
+		return node, nil
+	}
 	switch node.Kind {
 	case yaml.SequenceNode:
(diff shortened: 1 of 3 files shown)


laurazard:
This makes it read a bit weird, no? If `!reset` only removes and does not allow to redefine, having to write:
```yaml
foo: !reset null
```
is a bit weird, I would expect to just write
```yaml
foo: !reset
```

Can we assume `null` when there is no value for `!reset`? (ignore this if that's already something we can do)

ndeloof:
@laurazard I we one can, but `!reset` is just a yaml tag, yaml syntax requires a value. Can also use:
```yaml
foo: !reset {}
```

laurazard:
RIght, I had a vague idea from when I was looking into the tags there was some reason why it was required. Regardless, that's unrelated to this change, `!override` sounds good to me.

PR #553, opened 2024-01-30
Fix relative path resolution when combining include+extends

ndeloof:
relative references resolved by localResourceLoader must be relative to include's project directory, not caller project directory

fixes https://github.com/docker/compose/issues/11412
fixes https://github.com/docker/compose/issues/11444


glours (approved):
LGTM

PR #555, opened 2024-01-30
Support urls as config files

### LINKED ISSUE #554: Support URLs as config files
Related to docker/compose#10360 but I think it makes more sense here.

To make using remote docker compose config files more convenient it should be possible for the user to specify a URL instead of needing to CURL and pipe it in manually.

There are many issues and discussions surrounding adding the ability for users to specify a URL as a config file, however the approach of using the -f flag transparently as opposed to a different --url flag seems cleaner.

Mjaethers:
Closes #554.

When ingesting the config files if one is detected to be a URL instead of a file path a http get request is sent and the underlying file is read.
If the base file is a URL the working directory is set to the user working directory.

Tests that compare using a URL and using local files to create a project have been included.

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


ndeloof:
You should rely on `ResourceLoader` to support accessing a (remote) compose file over HTTP(s)

Mjaethers:
I think we may be misunderstanding each other. As far as I can tell a `ResourceLoader` is used to load a remote service specified within the local compose file. What I am trying to do is add a way of specifying a compose file when calling `docker compose -f <filename> build` so that `<filename>` can also be a URL.

ndeloof:
sure, but still: ResourceLoader is the way to load non-local files, and get a local (cached) copy. Support for `docker compose -f git://xxx` should rely on it for homogeneity with other places we support this feature

PR #556, opened 2024-01-31
fix cycle tracker to manage homonym services

ndeloof:
fix detection of extends cycles

closes https://github.com/docker/compose/issues/11430


glours (approved):
LGTM

PR #558, opened 2024-02-02
set default values to required attributes

ndeloof:
This is an alternative to https://github.com/compose-spec/compose-go/pull/557

having an explicit "set-defaults" step parsing compose.yaml files allows to keep the `extends` source in its original form, without an implicit `context: .` value, and prevent incorrect path resolution to take place depending we extends within same file (already parsed) or from another one (to be parsed).


glours (approved):
LGTM

PR #561, opened 2024-02-05
Add getters for services that have defined build, extends, depends_on, capabilities, gpu and tpu

jhrotko:
Add methods to retrieve which services have defined:
- build
- extends
- depends_on
- capabilities
- gpu
- tpu

This will be used in compose for two telemetry purposes. We want to know the complexity of compose files and if compose is being used for AI/ML development

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -29,6 +29,7 @@ import (
 	"github.com/distribution/reference"
 	"github.com/mitchellh/copystructure"
 	godigest "github.com/opencontainers/go-digest"
+	"golang.org/x/exp/maps"
 	"golang.org/x/sync/errgroup"
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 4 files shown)


types/project.go
```diff
@@ -119,6 +119,36 @@ func (p *Project) ConfigNames() []string {
 	return names
 }
 
+func (p *Project) ServicesBuild() []string {
```
ndeloof, line 122: For even more flexibility we could introduce `ServicesWith(predicate func(ServiceConfig) bool) []string`

By the way, makes me wonder we only want service names here, or actual services. Could have both by introducing 
`ServicesWith(predicate ...)` and `ServiceNamesWith(predicate ...)`, wdyt ?

jhrotko: fixed, could you please review?

ndeloof (approved):
LGMT
Wonder the naming would be clearer as `ServicesWithExtends()` ?

PR #563, opened 2024-02-05
host list can be used to declare both IPv4 and IPv6 for same hostname

ndeloof:
`HostList` was refactored as a `map[string]` but we missed the case where a single hostname can be mapped to multiple IPs, especially as one can have both IPv4 and IPv6 addresses


glours (approved):
LGTM to merge when CI will be happy

PR #564, opened 2024-02-06
introduce ability to register types for extensions

ndeloof:
This makes it easier to experiment with extensions and get data mapped into go structs


glours (approved):
🥳

PR #566, opened 2024-02-06
introduce port.name

ndeloof:
introduce port.name to document role of a port within a service


laurazard:
Maybe we can add a test for this/add it to one of the full parse tests?

ndeloof:
added a (dedicated) test - I prefer we avoid making the "full struct" test even more complex :)

ndeloof:
> Is the name an optional attribut, should we adapt the portsIndexer

no, this one is not part of the key, this is just metadata - while this probably should be unique, but this is only for user information, so \\_(ツ)_/

glours (approved):

PR #567, opened 2024-02-06
schema is hosted by compose-go, no need to sync anymore

ndeloof:
json schema is hosted in compose-go repository in sync with implementation code. This is now compose-spec doing the sync since https://github.com/compose-spec/compose-spec/pull/455


glours (approved):
My bad I forgot to remove this in my previous PR 🤦‍♂️ 
LGTM

PR #568, opened 2024-02-07
validate `extends` source does not share resources with another service/container

ndeloof:
applies https://github.com/compose-spec/compose-spec/blob/master/05-services.md#restrictions


glours (approved):
LGTM

jhrotko (approved):

PR #571, opened 2024-02-12
Add listeners to trace file metadata

jhrotko:
## Description
This PR intends to retrieve metadata from the parsing file. This was an issue when tracking the occurrences of `extends` in a project for telemetry since at the time the metrics are sent the Project is fully resolved thus no extends instances.

This will be one out of three PRs.
https://github.com/docker/compose/pull/11492
## Interaction

```mermaid
sequenceDiagram
    participant pinata
    participant compose
    participant compose-go
    compose->>compose-go: register listeners 
    compose->>+compose-go: load project
    compose-go->>+compose-go/extends: ApplyExtends
   Note right of compose-go/extends: options.ProcessEvent("extends", metadata)
    compose-go/extends-->>-compose-go: finish extends
    compose-go-->>-compose: finish load
    compose-->>pinata: send OTEL metrics
```
## Fix
by adding the listener to increment the number of processed extends, found a bug where when going down the search tree, we were not updating the leafs after the merge. In a case where we had chained extends, we would over look the tree, since we would only update the `services` at the head of the recursion in `ApplyExtends`

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -80,6 +80,10 @@ type ProjectOptions struct {
 	EnvFiles []string
 
 	loadOptions []func(*loader.Options)
+
+	// Callbacks to retrieve metadata information during parse defined before
+	// creating the project
+	Listeners []loader.Listener
 }
 
 type ProjectOptionsFn func(*ProjectOptions) error
@@ -89,6 +93,7 @@ func NewProjectOptions(configs []string, opts ...ProjectOptionsFn) (*ProjectOpti
 	options := &ProjectOptions{
 		ConfigPaths: configs,
 		Environment: map[string]string{},
+		Listeners:   []loader.Listener{},
 	}
 	for _, o := range opts {
 		err := o(options)
@@ -365,6 +370,11 @@ func WithExtension(name string, typ any) ProjectOptionsFn {
 	}
 }
 
+// Append listener to event
+func (o *ProjectOptions) WithListeners(listeners ...loader.Listener) {
+	o.Listeners = append(o.Listeners, listeners...)
+}
+
 // WithoutEnvironmentResolution disable environment resolution
 func WithoutEnvironmentResolution(o *ProjectOptions) error {
 	o.loadOptions = append(o.loadOptions, func(options *loader.Options) {
@@ -437,7 +447,8 @@ func ProjectFromOptions(options *ProjectOptions) (*types.Project, error) {
 
 	options.loadOptions = append(options.loadOptions,
 		withNamePrecedenceLoad(absWorkingDir, options),
-		withConvertWindowsPaths(options))
+		withConvertWindowsPaths(options),
+		withListener(options))
 
 	ctx := options.ctx
 	if ctx == nil {
@@ -477,6 +488,13 @@ func withConvertWindowsPaths(options *ProjectOptions) func(*loader.Options) {
 		if o.ResolvePaths {
 			o.ConvertWindowsPaths = utils.StringToBool(options.Environment["COMPOSE_CONVERT_WINDOWS_PATHS"])
 		}
+	}
+}
+
+// save listeners from ProjectOptions (compose) to loader.Options
+func withListener(options *ProjectOptions) func(*loader.Options) {
+	return func(opts *loader.Options) {
+		opts.Listeners = options.Listeners
 	}
 }
 
--- a/loader/extends.go
+++ b/loader/extends.go
@@ -70,8 +70,10 @@ func applyServiceExtends(ctx context.Context, name string, services map[string]a
 	case map[string]any:
 		ref = v["service"].(string)
 		file = v["file"]
+		opts.ProcessEvent("extends", v)
 	case string:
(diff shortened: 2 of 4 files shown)


cli/options.go
```diff
@@ -80,6 +80,10 @@ type ProjectOptions struct {
 	EnvFiles []string
 
 	loadOptions []func(*loader.Options)
+
+	// Callbacks to retrieve metadata information during parse defined before
+	// creating the project
+	Listeners map[string][]loader.Listener
```
ndeloof, line 86: This makes it a bit complicated to write a listener for more than one event type. Could declare `Listeners []Listener` as `Listener` is a `func(event string, details any`, it already receives the event type, and writing a listener then is just a question of a `switch` and can select any interesting event and ignore all others.

loader/extends.go
```diff
@@ -60,6 +60,8 @@ func applyServiceExtends(ctx context.Context, name string, services map[string]a
 	if !ok {
 		return s, nil
 	}
+	opts.ProcessEvent("extends", name)
```
ndeloof, line 63: in addition to service name, would be useful to track the file so we can detect all files involved in a compose model
`Listener`'s metadata parameter could be declared a `map[string]any` so it is easy to pass more than just one attribute here, and add more as this evolves without impacting consumers

loader/extends_test.go
```diff
@@ -60,11 +61,19 @@ services:
 			},
 		},
 		WorkingDir: abs,
+	}, func(options *Options) {
+		options.ResolvePaths = false
+		options.Listeners = map[string][]Listener{
+			"extends": {func(event string, metadata any) {
+				extendsCount++
+			}},
```
ndeloof, line 69:
```suggestion
                  options.Listeners = []Listeners{
                    func(event string, metadata map[string]any {
                       if event == "extends" {
                              count ++
                       }
			}},
```

ndeloof (changes requested):
Some design suggestion for more flexibility by consumer

Request changes on `Project`: it should NOT own FileMetadata, leave use of parsing events to the consumer - here you want to count extends, but there's other scenario we want to list files being processed, or ... mode details TBD

cli/options.go
jhrotko: @ndeloof by writing the way you are suggesting, you are passing the map functionality inside every registered listener by writing in all further listeners with a switch. All registered listeners need to define the switch. When in reality with the current solution we take advantage of the map data structure to handle this. Also, when processing 1 listener, with your recomendation, you will need to iterate over all listeners, when in reality you just care for a subset of them. I can see one advantage in having a `[]Listener` instead of `map[string][]Listener` which is, in case you want to have a particular Listener that is triggered for all events

loader/extends.go
jhrotko: this was a leftover from a test :) I intended to pass the `extends` (the whole extends metadata from the service) I will change this. I left a `any` in case for some reason in the future someone does not want to necessarily pass a `map[string]any`. What do you think @ndeloof ?

ndeloof: the benefit of a `map[string]any` is that this allows to "add more" without impact on previously reported keys. Using `any` the consumer would need to adjust to any change that's made to the original struct

milas (commented):
Agree with Nicolas' suggestions - I think a generic listener that gets the event type + metadata is fine, consumers can ignore what they don't care about. Other than that and last cleanup though, lgtm!

cli/options.go
```diff
@@ -365,6 +370,11 @@ func WithExtension(name string, typ any) ProjectOptionsFn {
 	}
 }
 
+// Append listener to event
+func (o *ProjectOptions) WithListener(listeners []loader.Listener) {
+	o.Listeners = listeners
```
ndeloof, line 375: nit: Better make this a appending function so it can be used multiple times
```suggestion
func (o *ProjectOptions) WithListener(listener loader.Listener) {
	o.Listeners = append(o.Listeners, listener)
```

ndeloof (approved):
LGTM

cli/options.go
jhrotko: https://github.com/compose-spec/compose-go/pull/571/commits/6e4fb2bfbf56a2bd449bd627c3d3469fd331da2c#diff-0953236283edf016dbef569cb026a05dba9df5e7934af11b82499e3bec8fe006R374

cli/options.go
```diff
@@ -480,6 +491,13 @@ func withConvertWindowsPaths(options *ProjectOptions) func(*loader.Options) {
 	}
 }
 
+// save listeners from ProjectOptions (compose) to loader.Options
+func withListener(options *ProjectOptions) func(*loader.Options) {
```
glours, line 495: As there could be more than one listener to save
```suggestion
func withListeners(options *ProjectOptions) func(*loader.Options) {
```

ndeloof: indeed. I plan to add `include` metric support using this mechanism so will fix this at same time in a follow-up PR

PR #572, opened 2024-02-13
expose implicit UID and GID variables

ndeloof:
A very common issue using compose is to bind mount local files into container, and get into permission issues. Until docker engine provides some uid+gid translation on bind mounts (maybe using bindfs?) the only option is to run container with same user as the owner of the bind-mounted file
This introduce support for `UID` and `GID` variables on systems this makes sense (i.e. everything but Windows) so one can use `user: ${UID}:${GID}` in a compose file.


milas (commented):
I can't find it right now but there has been a previous discussion about this - will look it up and link here later (on mobile right now)

I definitely get the desire to get the host UID/GID, particularly for bind mounts permissions, but I also am nervous about introducing something explicitly non-portable (and not even supported by all our platforms)

ndeloof:
Example request for such  feature: https://github.com/docker/compose/issues/4081#issuecomment-1937787722
I 100% agree not being totally confident with such a non-portable feature

PR #478, opened 2023-11-02
add support for inline config

ndeloof:
This adds support for inlined config (https://github.com/compose-spec/compose-spec/pull/429)

PR #573, opened 2024-02-14
Add inline content support for secrets

yp05327:
In #478, inlined content for config was supported.
`secrets` and `configs` are using the same struct `FileObjectConfig`, so it can be easily implemented.

When I want to manage all secrets in one compose file, I noticed that I need to manage both compose files and the secrets files, as secrets only support reading from a file or environment variables.


ndeloof:
The reason inlined content was introduced for configs and not for secrets is that in most scenario, a secret is local or dynamically generated and should not be committed with the main compose file. Also to be considered security risks for users misusing such a feature and actually pushing to github some real secrets (yes, this happens, and at large scale)
Other than such considerations, I don't see any blocker here

yp05327:
I understand the security considerations. It depends on the user's knowledge. But warnings can be added in docs.
Actually, there's no warnings about it in docs now. So still some users are pushing secrets without recognition.
In another way, for users who know how to use this feature correctly, they have more choices and can do more things.
I'm doing some works on docker compose based 'helm chart' which uses go template to generate docker compose files.
But I notice that users can only create a secret file first, then link it in docker compose.
In k8s, they provide such feature, and have security warnings about using secrets in docs.
So if this feature can be supported, it will be awesome.

ndeloof:
This feature would be useful as a workaround for https://github.com/docker/compose/issues/11941

ndeloof (approved):

yp05327:
As this pr is created 5 month ago, I updated the branch and let's check the CI's result again.

ndeloof:
After debating this with the team, we don't want to risk users would start using plain text secrets in their config file, even well documented. So I'm closing this PR

jvitor83:
@ndeloof this should be supported at least for `config`.
The `secret` use of `content` can still be blocked, but the `config` is a real and authentic way of use.
I think this can be reopened and changed to support at least for the config.

ndeloof:
config already has support for inlined content

PR #574, opened 2024-02-14
pass context as first parameter

ndeloof:
As compose/v2 introduces a few breaking changes, I suggest we also remove `Options.WithContext` and prefer use of `ctx` passed as first func argument, applying recommended go-idom


laurazard (approved):

glours (approved):
LGTM

PR #575, opened 2024-02-14
warn user `version` is obsolete

ndeloof:
Since compose-specification was released, `version` attribute is obsolete as it doesn't guarantee any form of compatibility. While users adopted new features introduced in the spec, we still see many of them using `version: "3"` in compose files.

This adds a warning to encourage users to stop using this useless attirbute

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -438,6 +438,11 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


glours (approved):

jhrotko (approved):

rfay:
I'm glad to see this :) SO many docker-compose.*.yaml add-ons for DDEV have `version: 3.6` in them :)

PR #578, opened 2024-02-16
ignore failure parsing yaml file when looking for project name

ndeloof:
postpone checking a project name is set, so that we can just ignore yaml parsing error looking for `name`

fixes https://github.com/docker/compose/issues/11450


glours (approved):
LGTM

PR #579, opened 2024-02-16
process include with a relative-path local resource loader

ndeloof:
As we process `include`, use relative path to configure local resource loader so that included resources point to the target path, but still respect the `--no-path-resolution` flag

fixes https://github.com/docker/compose/issues/11508


glours (approved):
LGTM

PR #582, opened 2024-02-19
Fix env file read from include

jhrotko:
When reading a custom env file from include, we were not passing the read enviroenmnt variables to the config Environments, thus the read vars were being "forgotten".
Also fixed the default directory for environment variables, which should be the same as the compose.yml that is invoking the `include`

`withListeners` change forgotten in previous [PR](https://github.com/compose-spec/compose-go/pull/571)

Fixes: 
https://github.com/docker/compose/issues/11509
https://docker.atlassian.net/browse/COMP-438

### DIFF
--- a/loader/include.go
+++ b/loader/include.go
@@ -104,6 +104,22 @@ func ApplyInclude(ctx context.Context, configDetails types.ConfigDetails, model 
 			if s, err := os.Stat(f); err == nil && !s.IsDir() {
 				r.EnvFile = types.StringList{f}
 			}
+		} else {
+			envFile := []string{}
+			for _, f := range r.EnvFile {
+				if !filepath.IsAbs(f) {
+					f = filepath.Join(configDetails.WorkingDir, f)
+					s, err := os.Stat(f)
+					if err != nil {
+						return err
+					}
+					if s.IsDir() {
+						return fmt.Errorf("%s is not a file", f)
+					}
+				}
+				envFile = append(envFile, f)
+			}
+			r.EnvFile = envFile
 		}
 
 		envFromFile, err := dotenv.GetEnvFromFile(configDetails.Environment, r.EnvFile)
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -453,6 +453,7 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 			return nil, err
 		}
 	}
+	resolveServicesEnvironment(dict, config)
 
 	return dict, nil
 }
--- a/cli/options.go
+++ b/cli/options.go
@@ -442,7 +442,7 @@ func ProjectFromOptions(ctx context.Context, options *ProjectOptions) (*types.Pr
 	options.loadOptions = append(options.loadOptions,
 		withNamePrecedenceLoad(workingDir, options),
 		withConvertWindowsPaths(options),
-		withListener(options))
+		withListeners(options))
 
 	project, err := loader.LoadWithContext(ctx, types.ConfigDetails{
 		ConfigFiles: configs,
@@ -484,9 +484,9 @@ func withConvertWindowsPaths(options *ProjectOptions) func(*loader.Options) {
 }
 
 // save listeners from ProjectOptions (compose) to loader.Options
-func withListener(options *ProjectOptions) func(*loader.Options) {
+func withListeners(options *ProjectOptions) func(*loader.Options) {
 	return func(opts *loader.Options) {
-		opts.Listeners = options.Listeners
+		opts.Listeners = append(opts.Listeners, options.Listeners...)
 	}
 }
 
--- a/loader/environment.go
+++ b/loader/environment.go
@@ -1,0 +1,59 @@
+/*
+   Copyright 2020 The Compose Specification Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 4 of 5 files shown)


loader/include.go
```diff
@@ -104,17 +104,27 @@ func ApplyInclude(ctx context.Context, configDetails types.ConfigDetails, model
 			if s, err := os.Stat(f); err == nil && !s.IsDir() {
 				r.EnvFile = types.StringList{f}
 			}
+		} else {
+			envFile := []string{}
+			for _, f := range r.EnvFile {
+				f := filepath.Join(configDetails.WorkingDir, f)
```
ndeloof, line 110: need to guard this with `filepath.IsAbs(r.Envfile)` as env_file can also be set to an absolute path, or relative to user's home `~`

loader/include.go
```diff
@@ -104,17 +104,27 @@ func ApplyInclude(ctx context.Context, configDetails types.ConfigDetails, model
...
+				if s, err := os.Stat(f); err == nil && !s.IsDir() {
```
ndeloof, line 111: if explicitly selected env file is a Dir, then this is an error. We only ignore when selecting the _default_ `.env` file.

loader/include.go
```diff
@@ -104,17 +104,27 @@ func ApplyInclude(ctx context.Context, configDetails types.ConfigDetails, model
...
+				}
+			}
+			r.EnvFile = envFile
 		}
 
 		envFromFile, err := dotenv.GetEnvFromFile(configDetails.Environment, r.EnvFile)
 		if err != nil {
 			return err
 		}
+		configDetails.Environment.Merge(envFromFile)
```
ndeloof, line 122: Why this change ?
Need to clone before merge, otherwise env_file selected to parse the included model will still be present in `configDetails` and will have impacts on next include being processed

jhrotko: It took me some time to understand what is the error. Currently, the only environment variable that are resolved DURING the parse is the Interpolation (`$VAR`) and this case `environment: - VAR` will only be solved AFTER the load in https://github.com/compose-spec/compose-go/blob/4c12ed69741da3fc0acecd36bf6504687a95517e/loader/loader.go#L528 
That's why changing the `Merge` to also be applied to the higher level `configDetails` works. However, this is not correct, the environment variables of `custom.env` (from the test and the issue example) should be only within the context of the `include`. I propose to apply a `resolveEnvironment` at the end of the `loadYamlModel` to have this case into consideration, without changing the overall `project.Environment`

loader/loader.go
```diff
@@ -787,3 +788,29 @@ func convertVolumePath(volume types.ServiceVolumeConfig) types.ServiceVolumeConf
 	volume.Source = convertedSource
 	return volume
 }
+
+func resolveEnvironment(dict map[string]any, opts *Options) {
```
ndeloof, line 792: I was confused reviewing this PR about this func name, which is close to `Project.ResolveEnvironment` but serve a very different scenario, but I don't have a better name for now 😅
Will create a follow-up PR if I find one

glours: Ok I'm not alone 😅

loader/loader.go
```diff
@@ -787,3 +788,29 @@ func convertVolumePath(volume types.ServiceVolumeConfig) types.ServiceVolumeConf
...
+	for key, value := range dict {
+		if key == "environment" {
+			if environment, ok := value.([]any); ok {
+				serviceEnvironment := []any{}
+				for _, env := range environment {
+					env, ok := env.(string)
+					if !ok {
+						continue
+					}
+					if found, ok := opts.Interpolate.LookupValue(env); ok {
```
ndeloof, line 802: `Interpolate` being involved here is a bit confusing. I understand the actual goal is to access environment map, which `LookupValue` captured in options. So this basically is a hack - We should better add an explicit `Environment` (or `Variables` ?) attribute to `loader.Options` so we make it explicit we rely on the `ConfigDetails.Environment` here (and maybe in other places in future)

loader/loader.go
```diff
@@ -787,3 +788,29 @@ func convertVolumePath(volume types.ServiceVolumeConfig) types.ServiceVolumeConf
...
+					}
+
+				}
+				dict["environment"] = serviceEnvironment
+			}
+		}
+		if v, ok := value.(map[string]any); ok {
+			resolveEnvironment(v, opts)
+		}
+	}
```
glours, line 815: We don't need to loop on the whole `dict` map IHMO
```suggestion
if value, ok := dict["environment"]; ok {
		if environment, ok := value.([]any); ok {
			serviceEnvironment := []any{}
			for _, env := range environment {
				env, ok := env.(string)
				if !ok {
					continue
				}
				if found, ok := opts.Interpolate.LookupValue(env); ok {
					serviceEnvironment = append(serviceEnvironment, fmt.Sprintf("%s=%s", env, found))
				} else {
					serviceEnvironment = append(serviceEnvironment, env)
				}

			}
			dict["environment"] = serviceEnvironment
		}
		if v, ok := value.(map[string]any); ok {
			resolveEnvironment(v, opts)
		}
	}
```

glours (commented):
I think we should directly get the value from the map and avoid looping the whole dict

loader/loader.go
ndeloof: can also do 
```golang
switch environment := value.(type) {
case []any:
   for _, env := range environment {   ... }
case maps[string]any:
  resolveEnvironment(v, opts)
```

ndeloof (approved):

glours (approved):
LGTM

PR #584, opened 2024-02-20
merge volumes/networks labels

ndeloof:
Add some missing merge rules

avoid panic reported on https://github.com/docker/compose/issues/11520 while the root cause might be unrelated to this PR


glours (approved):
LGTM

PR #585, opened 2024-02-20
check that a container_name is used only once across all services declarations

### LINKED ISSUE #583: Validation doesn't catch multiple services with the same container_name
The following configuration declares two services, each of which has the `container_name` specifies.
But those container names are the same.

```yaml
version: '3.4'
services:
  webservice1:
    container_name: foo
    image: nginx
  webservice2:
    container_name: foo
    image: nginx
```

I can run `docker compose up` with the following configuration.
But there will be an error at runtime from Docker,
so one of the two containers will not start.

Why doesn't compose validate this and issue an error before running any containers?

glours:
Fix #583

### DIFF
--- a/loader/validate.go
+++ b/loader/validate.go
@@ -29,6 +29,7 @@ import (
 
 // checkConsistency validate a compose model is consistent
 func checkConsistency(project *types.Project) error {
+	containerNames := map[string]string{}
 	for _, s := range project.Services {
 		if s.Build == nil && s.Image == "" {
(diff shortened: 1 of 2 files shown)


loader/validate.go
```diff
@@ -29,6 +29,7 @@ import (
 
 // checkConsistency validate a compose model is consistent
 func checkConsistency(project *types.Project) error {
+	containerNames := map[string]string{}
```
ndeloof, line 32: could use `utils.Set`, which offers `Has` method

glours: I can't keep the original service name with a Set only the `container_name` value

jhrotko (approved):

PR #592, opened 2024-02-29
Use ResourceLoader to ingest compose files

Mjaethers:
Related to PR #555 and Issue #554 

To achieve consistency in the way resources are loaded for docker compose the config files are now ingested via `ResourceLoader`. For local files the existing `localResourceLoader` is used.

To add new ways of loading compose files in the future a new `ResourceLoader` should be implemented and added to the project options.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -279,6 +279,48 @@ func parseYAML(decoder *yaml.Decoder) (map[string]interface{}, PostProcessor, er
 	return converted.(map[string]interface{}), &processor, nil
 }
 
+// LoadConfigFiles ingests config files with ResourceLoader and returns config details with paths to local copies
+func LoadConfigFiles(ctx context.Context, configFiles []string, options ...func(*Options)) (*types.ConfigDetails, error) {
+	if len(configFiles) < 1 {
+		return &types.ConfigDetails{}, errors.New("no files specified")
+	}
+
+	opts := &Options{}
+	config := &types.ConfigDetails{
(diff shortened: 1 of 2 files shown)


loader/loader.go
```diff
@@ -279,6 +279,48 @@ func parseYAML(decoder *yaml.Decoder) (map[string]interface{}, PostProcessor, er
...
+	opts.ResourceLoaders = append(opts.ResourceLoaders, localResourceLoader{})
+
+	for i, p := range configFiles {
+		for _, loader := range opts.ResourceLoaders {
+			if !loader.Accept(p) {
+				continue
+			}
+			local, err := loader.Load(ctx, p)
+			if err != nil {
+				continue
```
ndeloof, line 305: why ignore error ?

ndeloof (commented):
need to experiment a bit with it, but LGTM in general design

loader/loader.go
Mjaethers: My thinking was that if Load() failed it would make sense to simply keep trying with another ResourceLoader as it could be caused by an overzealous Accept() implementation, but I realize now that the way it is currently set up if all the Load()s fail there'll be no error thrown until it tries to read an empty compose.

glours:
Hello @Mjaethers 
Sorry for the long period without feedback 🙇 
Is it possible for you to resolve the conflict and rebase your branch with main?

glours:
#701  Merged

PR #595, opened 2024-03-12
normalization creates implicit `default` network only when used by some service

ndeloof:
`default` implicit network should not be introduced by normalization if not explicitly used by a service.

This let `normalizeNetworks` detect `default" network is used by some services and create this implicit network only if required


glours (approved):
LGTM

PR #598, opened 2024-03-15
only warn once about obsolete version

ndeloof:
with include/extends/overrides, a bunch of compose files can be involved during project loading. Warning user for each and every `version` we found would result in many lines without clear information on the culprit(s) and probably repeated such error (as multiple services with extends could load same file many times)

this collect the files seen with a `version` attribute and only warn user once, with file name

also disable this warning if loading has been configured to skip validation


glours (approved):
LGTM

PR #602, opened 2024-03-18
introduce CheckCycle to detect cycles in project's depends_on

ndeloof:
fix: cycle detection was ran for each and every service in project, not once per project 😅
also introduced a more explicit way to use the graph logic to search for cycles, vs running a no-op "in dependency order", where cycle detection is just an implementation detail

see https://github.com/docker/compose/issues/11437


glours (approved):
🙏

PR #606, opened 2024-03-20
run checkConsistency after profiles were applied

ndeloof:
run checkConsistency after profiles were applied. 
If we don't, some services may refer to dependencies disabled by profile

As a side effect, this also allows to (ab)use profiles to define same resource twice with distinct configuration:
close https://github.com/docker/compose/issues/11619.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -554,15 +554,15 @@ func modelToProject(dict map[string]interface{}, opts *Options, configDetails ty
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


luciangabor:
There is also a question of how far do you want to go with such checks ...

```console

$ docker compose -f - up --scale service=2 <<-'EOF'
services:
  service:
    image: nginx
    container_name: service
EOF
WARNING: The "service" service is using the custom container name "service". Docker requires each container to have a unique name. Remove the custom name to scale the service.

```

PR #609, opened 2024-03-22
checkConsistency must ignore disabled service if required=false

ndeloof:
after https://github.com/compose-spec/compose-go/pull/606 `checkConsistency` will fail if a service referred by `depends_on`  is disabled by profile (see https://github.com/docker/compose/blob/main/pkg/e2e/fixtures/dependencies/deps-not-required.yaml)


jhrotko (approved):
LGTM

PR #613, opened 2024-04-02
Only ignore parsing errors when used with no-interpolate

ndeloof:
Parsing error must be ignored when interpolation is disabled so we don't try to parse `${PORT}` as an IP address or port number, **but** when interpolation is enabled (loading into `Project`) we must not ignore errors, otherwise the final error is a very generic one

closes https://github.com/docker/compose/issues/11676


transform/ports_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/compose-spec/compose-go/v2/tree"
+	"gotest.tools/v3/assert"
+)
+
+func Test_transformPorts(t *testing.T) {
+	tests := []struct {
+		name    string
+		yaml    any
+		want    any
```
glours, line 31: Why not add `ignoreParseError` bool here and a test to check we don't generate error when set to true?

glours (approved):
LGTM

PR #614, opened 2024-04-02
fix: `include.project_directory` path resolution

glours:
Fixes https://github.com/docker/compose/issues/11626 and https://github.com/docker/compose/issues/11573

https://docker.atlassian.net/browse/COMP-496

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -148,13 +148,24 @@ func (l localResourceLoader) Load(_ context.Context, p string) (string, error) {
 	return l.abs(p), nil
 }
 
-func (l localResourceLoader) Dir(path string) string {
-	path = l.abs(filepath.Dir(path))
+func (l localResourceLoader) Dir(originalPath string) string {
+	path := l.abs(originalPath)
+	if !l.isDir(path) {
+		path = l.abs(filepath.Dir(originalPath))
+	}
 	rel, err := filepath.Rel(l.WorkingDir, path)
 	if err != nil {
 		return path
 	}
 	return rel
+}
+
+func (l localResourceLoader) isDir(path string) bool {
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+	return fileInfo.IsDir()
 }
 
 func (o *Options) clone() *Options {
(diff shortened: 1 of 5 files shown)


ndeloof (approved):
🥳

loader/loader.go
```diff
@@ -148,15 +148,26 @@ func (l localResourceLoader) Load(_ context.Context, p string) (string, error) {
 	return l.abs(p), nil
 }
 
-func (l localResourceLoader) Dir(path string) string {
-	path = l.abs(filepath.Dir(path))
+func (l localResourceLoader) Dir(originalPath string) string {
+	path := l.abs(originalPath)
+	if !l.isDir(path) {
+		path = l.abs(filepath.Dir(originalPath))
+	}
```
milas, line 155: This heuristic makes me a bit nervous. We've had issues before doing stuff like this - the `isDir` check is doing a stat, so requires all the paths to exist to behave correctly. In this instance, that might be ok, but I'm not sure it holds for the general case and might result in surprises for future callers of `Dir()`.

In practice, while a pain, I think it'd be better to treat this more like the Go stdlib and ALWAYS return the enclosing dir and make the `Abs` function public (could rename it something like `LoadedPath()`?). That is, the caller needs to know what type of path they're carrying around (file or dir) and invoke the right function.

milas (approved):
Fix makes sense - logic to detect directories makes me nervous because it does a `stat` / real FS operation, which has bit us in the past

loader/loader.go
ndeloof: Can be checked running `compose-spec` command with `--no-path-resolution`

ndeloof: by the way, I'm confused why such a distinction is made. `Dir` should return owner directory, whenever path is a file or not

glours: depending how you declare your path `foo/bar` or `foor/bar/`, `filepath.Dir` won't return the same output, `foo` in the first case and `foo/bar` in the second one
https://pkg.go.dev/path/filepath#example-Dir

ndeloof: then AFAICT we could just rely on `filepath.Clean` so that we don't need to check actual file exists but rely on file path representation

PR #618, opened 2024-04-11
refs issue:11706 Resolving the Alias value of Alias type node

arankeparth:
**Changes:**

1.  If the node is of `Alias` kind, we need to preserve the current path and process the Alias value separately.
2. By doing this, the `!overwrite ` tag won't be ignored.

**Testing:**

1. Created setup that mentioned on this issue: https://github.com/docker/compose/issues/11706
2. Results:

containers are spawned successfully:

Here, we can see in the configs section for `y` that there is only one entry for `configs`

### DIFF
--- a/loader/reset.go
+++ b/loader/reset.go
@@ -42,14 +42,8 @@ func (p *ResetProcessor) UnmarshalYAML(value *yaml.Node) error {
 // resolveReset detects `!reset` tag being set on yaml nodes and record position in the yaml tree
 func (p *ResetProcessor) resolveReset(node *yaml.Node, path tree.Path) (*yaml.Node, error) {
 	// If the path contains "<<", removing the "<<" element and merging the path
-	if strings.Contains(path.String(), "<<") {
-		pathArr := strings.Split(path.String(), ".")
-		path = tree.NewPath(pathArr[0])
-		for _, el := range pathArr[1:] {
-			if el != "<<" {
-				path = tree.Path(strings.Join([]string{path.String(), el}, "."))
-			}
-		}
+	if strings.Contains(path.String(), ".<<") {
+		path = tree.NewPath(strings.Replace(path.String(), ".<<", "", 1))
 	}
 	// If the node is an alias, We need to process the alias field in order to consider the !override and !reset tags
 	if node.Kind == yaml.AliasNode {
--- a/loader/reset.go
+++ b/loader/reset.go
@@ -19,6 +19,7 @@ package loader
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/compose-spec/compose-go/v2/tree"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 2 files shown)


arankeparth:
Please review somebody

arankeparth:
@ndeloof please review.

loader/reset.go
```diff
@@ -40,6 +41,21 @@ func (p *ResetProcessor) UnmarshalYAML(value *yaml.Node) error {
 
 // resolveReset detects `!reset` tag being set on yaml nodes and record position in the yaml tree
 func (p *ResetProcessor) resolveReset(node *yaml.Node, path tree.Path) (*yaml.Node, error) {
+	// If the path contains "<<", removing the "<<" element and merging the path
+	if strings.Contains(path.String(), "<<") {
+		pathArr := strings.Split(path.String(), ".")
```
ndeloof, line 46: use path.Parts()

loader/reset.go
```diff
@@ -40,6 +41,21 @@ func (p *ResetProcessor) UnmarshalYAML(value *yaml.Node) error {
...
+		path = tree.NewPath(pathArr[0])
+		for _, el := range pathArr[1:] {
+			if el != "<<" {
+				path = tree.Path(strings.Join([]string{path.String(), el}, "."))
```
ndeloof, line 50: path = path.Next(el)

loader/reset.go
```diff
@@ -40,6 +41,21 @@ func (p *ResetProcessor) UnmarshalYAML(value *yaml.Node) error {
 
 // resolveReset detects `!reset` tag being set on yaml nodes and record position in the yaml tree
 func (p *ResetProcessor) resolveReset(node *yaml.Node, path tree.Path) (*yaml.Node, error) {
+	// If the path contains "<<", removing the "<<" element and merging the path
+	if strings.Contains(path.String(), "<<") {
```
ndeloof, line 45: isn't this just equivalent to `path = strings.Replace(path, ".<<", "")` ?

ndeloof (commented):
Can you please add a test case to demonstrate this actually fixes the merge issue?

loader/reset.go
arankeparth: Yes, I have made this change.

ndeloof (approved):

arankeparth:
@ndeloof how to get this merged?

ndeloof:
@arankeparth let's wait for a second review, then I think we can merge

arankeparth:
@glours please review.

arankeparth:
@milas can you please review this?

ndeloof:
@arankeparth please don't push pressure on the team, will be reviewed once we get bandwidth to review

arankeparth:
> @arankeparth please don't push pressure on the team, will be reviewed once we get bandwidth to review

Sorry for that.

glours (approved):

PR #619, opened 2024-04-11
remove constraint for container_name unicity while loading model

ndeloof:
such validation should occur later, as we actually select containers to be created

see https://github.com/docker/compose/issues/11619#issuecomment-2047995758 for context


arunk-r (approved):

glours (approved):
LGTM

milas (approved):
It's unfortunate to lose this validation here but I don't see a better alternative right now :(

PR #620, opened 2024-04-19
detect self-reference and better report depends_on cycles

ndeloof:
fixes https://github.com/docker/compose/issues/11586


glours (approved):
LGTM

PR #621, opened 2024-04-26
Fix env parsing panics

glours:
Return errors instead of trigger panic when asserting types.

Fixes https://github.com/docker/compose/issues/11772 and https://github.com/docker/compose/issues/11771

### DIFF
--- a/override/uncity.go
+++ b/override/uncity.go
@@ -198,12 +198,15 @@ func portIndexer(y any, p tree.Path) (string, error) {
 	return "", nil
 }
 
-func envFileIndexer(y any, _ tree.Path) (string, error) {
+func envFileIndexer(y any, p tree.Path) (string, error) {
 	switch value := y.(type) {
 	case string:
 		return value, nil
 	case map[string]any:
-		return value["path"].(string), nil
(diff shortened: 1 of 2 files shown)


laurazard (approved):
LGTM

ndeloof (approved):

override/uncity.go
```diff
@@ -198,12 +198,15 @@ func portIndexer(y any, p tree.Path) (string, error) {
...
+func envFileIndexer(y any, p tree.Path) (string, error) {
 	switch value := y.(type) {
 	case string:
 		return value, nil
 	case map[string]any:
-		return value["path"].(string), nil
+		if pathValue, ok := value["path"]; ok {
+			return pathValue.(string), nil
+		}
+		return "", fmt.Errorf("environment path attribut %s is missing", p)
```
crazy-max, line 209: `attribut > attribute` Just catched this typo while looking at 2.1.1 release.

ndeloof: you missed an opportunity to be recognized as contributor for next release :)
https://github.com/compose-spec/compose-go/pull/633

PR #625, opened 2024-05-13
Fix compose schema

jhrotko:
This PR replaces [gojsonschema](https://github.com/xeipuuv/gojsonschema) with [jsonschema](https://github.com/santhosh-tekuri/jsonschema) since the first does not validate correctly our schema (if at all), leading to unvalidated  compose-spec.json schema. If you open VSCode, and write the example
```
services:
  web:
    image: alpine
include:
  - some/path <<
```
VSCode (and other IDEs) will complain about `some/path`.

Some tests were introduced, but nearly not enough to fully cover the current compose schema.

Also, changed from pseudo draft2019 to draft07 since this is the most adopted draft spec.


ndeloof (approved):
🥳

loader/loader.go
```diff
@@ -309,7 +309,13 @@ func Load(configDetails types.ConfigDetails, options ...func(*Options)) (*types.
 // LoadWithContext reads a ConfigDetails and returns a fully loaded configuration as a compose-go Project
 func LoadWithContext(ctx context.Context, configDetails types.ConfigDetails, options ...func(*Options)) (*types.Project, error) {
 	opts := toOptions(&configDetails, options)
-	dict, err := loadModelWithContext(ctx, &configDetails, opts)
+	sch, err := schema.CreateComposeSchema()
+	if err != nil {
+		return nil, err
+	}
+	fctx := context.WithValue(ctx, consts.ComposeSchema{}, sch)
```
glours, line 316: Is there any specific reason to initialise the Json Schema parser here and not keep the same logic as previously? 🤔

loader/loader.go
```diff
@@ -427,12 +433,17 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
...
+				if !ok {
+					return fmt.Errorf("could not validate yaml: %s", err)
+				} else {
+					if err := sch.Validate(dict); err != nil {
+						return fmt.Errorf("validating %s: %w", file.Filename, err)
+					}
+					if _, ok := dict["version"]; ok {
+						opts.warnObsoleteVersion(file.Filename)
+						delete(dict, "version")
+					}
```
glours, line 446: Same this should not have change as we only switch the Json Schema library, changes  should only apply in the `schema.go` file, no?

glours (commented):
Not blocking for me, but I wonder why changing the library to parse the Json input results in modifying logic in `loader.go` file, maybe I'm missing something 🤔

loader/loader.go
ndeloof: This allows to only compile the schema once (stored in context) vs for each and every compose file we parse. This is indeed not required but a reasonable optimization afaict

loader/loader.go
jhrotko: I stored the schema in the context to compile it only once. That's the only change that is reflected in more lines due to the go boiler plate to get it from the context 
```
sch, ok := ctx.Value(consts.ComposeSchema{}).(schema.ComposeSchema)
!ok {
  return fmt.Errorf("could not validate yaml: %s", err)
} else {
  ...
}
```

loader/loader.go
jhrotko: Alternatively I think it's also valid to move the initialisation to `loadYamlModel` and remove it from the context

glours: Make sense, I missed the compilation have to be done each time we parse a file 👌

glours (approved):

PR #626, opened 2024-05-14
Fix variable extraction when string uses variables concatenation

ndeloof:
by using `(.*)` to capture default value, variable regex can't distinguish multiple variables being concatenated
trying to fix the regex doesn't seem trivial (at least for me) and introduces a risk to break variable interpolation, so I preferred to replicate the interpolate login in variable extraction to recursively search for variables in the default value

also moved variable extraction logic to a dedicated go file for clarity

fixes https://github.com/docker/compose/issues/11785


glours (approved):
LGTM

PR #627, opened 2024-05-14
Add driver_opts to services.networks

robmry:
It's possible to set driver options when creating a network using compose ... [compose spec](https://github.com/robmry/compose-go/blob/b08e2159a8b1fd78444c2ef434bd4913d74697d1/schema/compose-spec.json#L637-L641) / [engine API](https://github.com/moby/moby/blob/9d07820b221db010bf1bdc26ca904468804ca712/api/swagger.yaml#L10134-L10138).

But, not when connecting an endpoint to a network ... [compose spec](https://github.com/robmry/compose-go/blob/b08e2159a8b1fd78444c2ef434bd4913d74697d1/schema/compose-spec.json#L293-L300) / [engine API](https://github.com/moby/moby/blob/9d07820b221db010bf1bdc26ca904468804ca712/api/swagger.yaml#L2541-L2544).

The API field isn't new, but upcoming change https://github.com/moby/moby/pull/47686 means it'll be useful for setting per-interface sysctls.

This PR adds the field.
- A corresponding change will be needed in `compose-spec`, and
- `compose` will need an update to copy the driver options into the engine API's `EndpointSettings`.

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -1727,6 +1727,49 @@ networks:
 	assert.DeepEqual(t, config, expected, cmpopts.EquateEmpty())
 }
 
+func TestLoadServiceNetworkDriverOpts(t *testing.T) {
+	config, err := loadYAML(`
+name: load-service-network-driver-opts
+services:
+  foo:
+    image: alpine
+    networks:
+      network1:
+        driver_opts:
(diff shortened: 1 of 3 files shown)


ndeloof (approved):

glours (approved):
LGTM

PR #628, opened 2024-05-15
Always set COMPOSE_PROJECT_NAME

ndeloof:
Always set COMPOSE_PROJECT_NAME based on project name being computed from various options.
This enforce COMPOSE_PROJECT_NAME interpolation always get the expected value

fixes https://github.com/docker/compose/issues/11734


glours (approved):
LGTM

PR #630, opened 2024-05-16
Revert "Fix compose schema"

ndeloof:
Reverts compose-spec/compose-go#625

see https://github.com/compose-spec/compose-spec/issues/501

#625 adds a bunch of regexp to validate formats. We also have to support variable being used as values for interpolation, so can't add such a validation in the spec


glours (approved):
LGTM

PR #589, opened 2024-02-22
implement normalization as a yaml transformation

ndeloof:
This re-implement "normalization" of the compose model as a plain yaml tree (map[string]any) not relying on compose-go structs conversion. This will make it possible to run the full compose processing (i.e `docker compose config`) without using go bindings, and then restore support for `--no-interpolate` and such

PR #634, opened 2024-05-31
Add option to keep empty environment variable keys in normalization

edwardrf:
In #589, the normalisation logic strips off any environment variables that does not have a value in config, however, i my use case, I am still interested in any empty environment variables names that has been set, which i'll resolve the value at a later stage, so I am implementing an naive option to allow me keep them during normalisation, please let me know if there is a better way of doing it. Currently, I have to load the project twice, use the one without normalisation to detect those empty environment variables.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -67,6 +67,8 @@ type Options struct {
 	SkipResolveEnvironment bool
 	// SkipDefaultValues will ignore missing required attributes
 	SkipDefaultValues bool
+	// KeepEmptyEnvironment will keep empty environment variable keys in the environment section
+	KeepEmptyEnvironment bool
 	// Interpolation options
 	Interpolate *interp.Options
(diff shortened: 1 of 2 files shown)


ndeloof:
please clarify your use case. a key-only environment variable is resolved based on user's environment, and ignored otherwise. Why not just populate the project environment with all usable values ?

edwardrf:
Any suggestions? I am open to an alternative approach if there is a recommendation.

ndeloof:
Actually this is a bug to remove empty variable from container environment, as empty has a special meaning to "remove" variable (see https://github.com/docker/compose/issues/11962)

ndeloof:
Closing as https://github.com/compose-spec/compose-go/pull/654 has been merged

edwardrf:
Thank you!

PR #635, opened 2024-06-03
go.mod: migrate to github.com/go-viper/mapstructure/v2 v2.0.0

thaJeztah:
- relates to https://github.com/mitchellh/mapstructure/issues/349
- relates to https://github.com/docker/cli/pull/4744

github.com/mitchellh/mapstructure will no longer be maintained by the author, and github.com/go-viper/mapstructure is nominated as the endorsed fork.

- v1.x changes since last release from mitchellh: https://github.com/go-viper/mapstructure/compare/v1.5.0...v1.6.0
- v2.0 changes: https://github.com/go-viper/mapstructure/compare/v1.6.0...v2.0.0

Breaking changes

Error is removed in favor of errors.Join (backported from Go 1.20 to preserve compatibility with earlier versions)

What's Changed

- feat!: update module path
- build: update dev env
- feature: add StringToBasicTypeHookFunc and support complex
- Add an example showing how to use a DecodeHookFunc to parse a custom field.
- Remove exposed error type
- Replace internal joined error with errors.Join

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,10 +6,10 @@ require (
 	github.com/distribution/reference v0.5.0
 	github.com/docker/go-connections v0.4.0
 	github.com/docker/go-units v0.5.0
+	github.com/go-viper/mapstructure/v2 v2.0.0
 	github.com/google/go-cmp v0.5.9
 	github.com/mattn/go-shellwords v1.0.12
 	github.com/mitchellh/copystructure v1.2.0
-	github.com/mitchellh/mapstructure v1.5.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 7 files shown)


ndeloof (approved):

ndeloof:
Nice!
I was worried with this dependency being unmaintained.

That being said, with compose-go to be able to process _most_ of the parse/interpolate/merge process in plain yaml, I think we could eventually get rid of this library by marshaling back the processed compose model into yaml and load it into go structs with a plain `yaml.Unmarshal`. Doing so we would have a more explicit "pre-processing" phase (shared with `docker compose config`) then load into go structs, which should be considered an "implementation detail"

thaJeztah:
> That being said, with compose-go to be able to process most of the parse/interpolate/merge process in plain yaml, I think we could eventually get rid of this library

I haven't looked closely where it's all used, but yes, if we can get rid of a dependency altogether, that's always good!

I think I noticed this one being mentioned when Mitchell announced he'd be deprecating various of his repositories, and I had a draft in docker/cli for some time. Now that the fork did a v2.0.0 (non-beta) release, I went looking where else it was used, and found it was used here :smile:

laurazard (approved):
LGTM

PR #637, opened 2024-06-07
networks.enable_ipv6: make it optional

akerouanton:
This field isn't marked as required in the compose-spec, but it's implemented as a basic `bool` in this package. This means it automatically defaults to `false` when not specified.

A recent change was made to Docker Engine to allow users to define a daemon-wide default value for the equivalent API field. Without this change, Compose would always send a value to the Engine API, preventing it from defining its own default value.

See https://github.com/moby/moby/pull/47867.

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -1689,6 +1689,7 @@ networks:
 
 	workingDir, err := os.Getwd()
 	assert.NilError(t, err)
+	enableIPv6 := true
 	expected := &types.Project{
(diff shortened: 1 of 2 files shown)


laurazard (approved):
LGTM

ndeloof:
LGTM as this reflects change in the API https://github.com/moby/moby/pull/47867/files#diff-28bc7f6688f2249afb33f1c3b4558c3bf1bfcda87f1b8e385bc8a5f04b8bdfeaL454-R454

PR #640, opened 2024-06-18
inform rather than warn on version obsolescence

rseymour:
docker compose warning on the version being obsolete has caused a lot of grief. Here is one simple solution to the issue.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -85,13 +85,13 @@ type Options struct {
 	Listeners []Listener
 }
 
-var versionWarning []string
+var versionObsolete []string
 
-func (o *Options) warnObsoleteVersion(file string) {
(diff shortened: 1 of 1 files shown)


ndeloof:
> docker compose warning on the version being obsolete has caused a lot of grief

no issue has been reported regarding this, so I don't think so

> Here is one simple solution to the issue.

:-1 this would basically let user continue setting `version` and expect this has some effect, which brings confusion. Compose spec was announced in 2020 with `version` made obsolete, it's time to move forward

glours:
I agree with @ndeloof, if we keep this as a simple message users will continue to use it and being confused when they will open issues about the file version v2 vs v3, and even sometimes with the Compose binary version itself. 
A warning message make it clear, it supposes you'll take action on your side if you don't want to be bother by the message again

PR #642, opened 2024-06-20
improve warning message about version attribute

### LINKED ISSUE #641: Version warning message causing lots of churn amongst projects with minimal benefit
I see https://github.com/compose-spec/compose-go/pull/640 was just closed due to lack of a raised issue as I was typing this comment there, so I'm raising the issue for discussion first.

There is a bunch of discussion about the impact of the warning and the questionable benefits in https://github.com/docker/compose/issues/11628. It's likely that the majority of users of `docker-compose` are not aware that the compose spec is a separate project - I wasn't until just now when I went to look at the history of this change.

I can see that the warning was first introduced in https://github.com/compose-spec/compose-go/pull/575 and later added to the spec in https://github.com/compose-spec/compose-spec/pull/489 - but it's unclear from those PRs how much discussion was had in other forums and whether the cost of creating warnings outweights the benefits vs simply ignoring the field.

These comments summarise the cost of introducing the warning and showing it to loads of people:
https://github.com/docker/compose/issues/11628#issuecomment-2061299025
https://github.com/docker/compose/issues/11628#issuecomment-2176213424
https://github.com/docker/compose/issues/11628#issuecomment-2176316608

Are there large costs to retaining the `version` field but ignoring it that I've overlooked?

glours:
fix #641

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -89,7 +89,7 @@ var versionWarning []string
 
 func (o *Options) warnObsoleteVersion(file string) {
 	if !slices.Contains(versionWarning, file) {
-		logrus.Warning(fmt.Sprintf("%s: `version` is obsolete", file))
+		logrus.Warning(fmt.Sprintf("%s: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion", file))
 	}
 	versionWarning = append(versionWarning, file)
 }


loader/loader.go
```diff
@@ -89,7 +89,7 @@ var versionWarning []string
 
 func (o *Options) warnObsoleteVersion(file string) {
 	if !slices.Contains(versionWarning, file) {
-		logrus.Warning(fmt.Sprintf("%s: `version` is obsolete", file))
+		logrus.Warning(fmt.Sprintf("%s: the attribute `version` is deprecated,it will be ignored, please remove it to avoid potential confusion", file))
```
ndeloof, line 92: version is not deprecated. If we want to, need first to update https://github.com/compose-spec/compose-spec/blob/master/04-version-and-name.md.

glours: thanks, updated

ndeloof (approved):

loader/loader.go
```diff
@@ -89,7 +89,7 @@ var versionWarning []string
 
 func (o *Options) warnObsoleteVersion(file string) {
 	if !slices.Contains(versionWarning, file) {
-		logrus.Warning(fmt.Sprintf("%s: `version` is obsolete", file))
+		logrus.Warning(fmt.Sprintf("%s: the attribute `version` is obsolete,it will be ignored, please remove it to avoid potential confusion", file))
```
ndeloof, line 92: obsolete,it => obsolete, it

PR #644, opened 2024-06-23
feat(loader): made interpolation after merge

idsulik:
**What this PR does / why we need it**:
Changed interpolation order, it used to interpolate for each config file separately, but now it interpolates after merge.  
To be honest, I'm not sure it's a good idea, maybe it'll be better to add an option to control it, like `--interpolate-after-merge`

**Which issue(s) this PR fixes**:
Fixes https://github.com/docker/compose/issues/11925

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -358,6 +358,12 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 		dict = map[string]interface{}{}
 		err  error
 	)
+	var interpolateOptsBeforeMerge interp.Options
+
+	if opts.Interpolate != nil && !opts.SkipInterpolation {
+		interpolateOptsBeforeMerge = opts.Interpolate.Clone()
+		interpolateOptsBeforeMerge.KeysToInterpolate = []string{"include", "extends"}
+	}
 	for _, file := range config.ConfigFiles {
 		fctx := context.WithValue(ctx, consts.ComposeFileKey{}, file.Filename)
 		if file.Content == nil && file.Config == nil {
@@ -376,6 +382,13 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 			cfg, ok := converted.(map[string]interface{})
 			if !ok {
 				return errors.New("Top-level object must be a mapping")
+			}
+
+			if opts.Interpolate != nil && !opts.SkipInterpolation {
+				cfg, err = interp.Interpolate(cfg, interpolateOptsBeforeMerge)
+				if err != nil {
+					return err
+				}
 			}
 
 			fixEmptyNotNull(cfg)
@@ -409,13 +422,6 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 			dict, err = override.EnforceUnicity(dict)
 			if err != nil {
 				return err
-			}
-
-			if !opts.SkipValidation {
-				if _, ok := dict["version"]; ok {
-					opts.warnObsoleteVersion(file.Filename)
-					delete(dict, "version")
-				}
 			}
 
 			return err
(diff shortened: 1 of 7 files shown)


ndeloof:
This was my initial intent during compose-go/v2 refactoring, but this breaks when `include` or `extends` uses variables for target file/service. Need _at least_ partial interpolation for those - I never took time to look into this, would you like to give it a try ?

idsulik:
> but this breaks when `include` or `extends` uses variables for target file/service

could you please provide an example?

ndeloof:
A typical use:
```yaml
services:
  prod:
    image: a

  dev:
    image: a-dev

  app:
    extends: ${BASE:-prod}
```

idsulik:
@ndeloof pushed changes to interpolate `include` and `extends` before merging and interpolate everything after the merge, but I think it becomes too complicated.

loader/loader.go
```diff
@@ -418,16 +424,6 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 				return err
 			}
 
-			if !opts.SkipValidation {
```
ndeloof, line 421: while remove per-file validation ?

loader/loader.go
```diff
@@ -358,6 +358,12 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 		dict = map[string]interface{}{}
 		err  error
 	)
+	var interpolateOptsBeforeMerge interp.Options
+
+	if opts.Interpolate != nil && !opts.SkipInterpolation {
+		interpolateOptsBeforeMerge = opts.Interpolate.Clone()
+		interpolateOptsBeforeMerge.KeysToInterpolate = []string{"include", "extends"}
```
ndeloof, line 365: AFAICT this would interpolate `include` but will skip `include.path`, won't it?
This deserves a dedicated test as there's a high risk to introduce a regression with such a change
I also think it would be safer to filter by path (`services.*.extends.**`)  vs attribute name

loader/loader.go
idsulik: accidentally removed, reverted it back

loader/loader.go
idsulik: Yes, it will. Fixed it, now it checks if the Path contains a given part.  
Before writing tests I want to know do we agree to do it this way?

tree/path.go
```diff
@@ -51,6 +51,16 @@ func (p Path) Parts() []string {
 	return strings.Split(string(p), pathSeparator)
 }
 
+func (p Path) ContainsPart(part string) bool {
```
ndeloof, line 54: This could work for this use case but is a bit fragile.
IMHO we could make this more deterministic if `Matches` adds support for "anything and child" `**` syntax, so that you can use :
```
KeysToInterpolate := []{
  "services.*.extends.**"
  "includes.**"
}
```

loader/loader.go
```diff
@@ -379,7 +385,7 @@ func loadYamlModel(ctx context.Context, config types.ConfigDetails, opts *Option
 			}
 
 			if opts.Interpolate != nil && !opts.SkipInterpolation {
-				cfg, err = interp.Interpolate(cfg, *opts.Interpolate)
+				cfg, err = interp.Interpolate(cfg, interpolateOptsBeforeMerge)
```
ndeloof, line 388: I wonder:
as extends will always require interpolation, wouldn't it be simpler to just force interpolation in `ApplyExtends` and `ApplyInclude`?

idsulik: 1. We call `ApplyExtends` only if `SkipExtends` is not `true` while interpolation happens always(the same for Include)
2. We call `ApplyInclude` after interpolation and `processors`, if we start interpolate inside `ApplyInclude` in can have consequences, I wanted to keep current logic as much as I can

tree/path.go
idsulik: pushed changes, wrote tests, if you know any other cases that can be useful, let me know

ndeloof:
Checking this PR again I went with an alternate implementation, reusing your test case: https://github.com/compose-spec/compose-go/pull/651

wdyt ?

ndeloof:
Closing as https://github.com/compose-spec/compose-go/pull/651 has been merged

PR #646, opened 2024-06-24
feat(options): Add WithUser to set current user id and group id

idsulik:
Issue https://github.com/docker/compose/issues/7853

Added `WithUserEnv` to support `COMPOSE_USER_ID` and `COMPOSE_GROUP_ID` env values in `compose.yaml`

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -86,6 +86,11 @@ type ProjectOptions struct {
 
 type ProjectOptionsFn func(*ProjectOptions) error
 
+var (
+	getUid = os.Getuid
+	getGid = os.Getgid
+)
+
 // NewProjectOptions creates ProjectOptions
 func NewProjectOptions(configs []string, opts ...ProjectOptionsFn) (*ProjectOptions, error) {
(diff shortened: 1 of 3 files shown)


ndeloof:
already discussed on https://github.com/compose-spec/compose-go/pull/572
no consensus here on the right way to address this issue

idsulik:
@ndeloof , oh, haven't noticed it. https://github.com/compose-spec/compose-spec/pull/282#pullrequestreview-1382209783 - what if I add it to compose repo? 

like that:
`cmd/compose/compose.go:353`
image

idsulik:
it's obvious that users need it according to comments and reactions https://github.com/docker/compose/issues/7853

ndeloof:
it's obvious there is an issue with bind mount ownership when it comes to local UID vs container's user, but this doesn't mean we all agree on the right way to manage this challenge. Adopting such UID/GID variables would makes compose configuration non-portable, which would break a major break in compose history after 10 years

PR #652, opened 2024-07-01
attributes also can be declared as string (using variables)

ndeloof:
This removes constraints on compose.yaml attributes as those can be set by variables, so we can't require value to be a number or format to be a duration, as `${VARIABLE}` also is a valid value

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -13,7 +13,6 @@
 
     "name": {
       "type": "string",
-      "pattern": "^[a-z0-9][a-z0-9_-]*$",
       "description": "define the Compose project name, until user defines one explicitly."
     },
 
@@ -94,7 +93,7 @@
         "develop": {"$ref": "#/definitions/development"},
         "deploy": {"$ref": "#/definitions/deployment"},
         "annotations": {"$ref": "#/definitions/list_or_dict"},
-        "attach": {"type": "boolean"},
+        "attach": {"type": ["boolean", "string"]},
         "build": {
           "oneOf": [
             {"type": "string"},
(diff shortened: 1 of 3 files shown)


schema/compose-spec.json
```diff
@@ -13,7 +13,6 @@
 
     "name": {
       "type": "string",
-      "pattern": "^[a-z0-9][a-z0-9_-]*$",
```
glours, line 16: Do we have this validation somewhere else in the source code?

ndeloof: yes, [NormalizeProjectName](https://github.com/compose-spec/compose-go/blob/main/loader/loader.go#L669-L673)

glours (approved):
LGTM

schema/compose-spec.json
```diff
@@ -94,7 +93,7 @@
         "develop": {"$ref": "#/definitions/development"},
         "deploy": {"$ref": "#/definitions/deployment"},
         "annotations": {"$ref": "#/definitions/list_or_dict"},
-        "attach": {"type": "boolean"},
+        "attach": {"type": ["boolean", "string"]},
```
jhrotko, line 96: should we create a type variable and add to all of these? a simple string is to relaxed of a restriction for this case

ndeloof: this schema is used 
1. by IDEs to drive autocompletion and validation, and as such must allow use of strings so `${VARIABLE}` is considered valid (and all variation of compose variable syntax)
2. in compose-go to check tree structure, but not node values. Failure to convert to go types is detected during unmarshalling

jhrotko: what I am saying is that with the current implementation 'banana' is a valid value for all these changes. Not necessarily the var syntax that we intend

ndeloof: sure, but such validation (which would require a complex regexp to express the variable syntax) would bring no benefit, as IDE won't use this for advanced validation AFAICT, and compose-go parser would anyway report an error

jhrotko: my suggestion would be to create a definition for the `${VARIABLE}` which should be like 
```
    "variable_type": {
      "type": "string",
      "pattern":"/^\$\{[a-zA-Z_$][a-zA-Z_$0-9]*\}$",
      "uniqueItems": true
    },
```
assuming we want 
`${VAR}`
`${hehe_123}` 
`${_123}` 
`${123}` -> no
`${123yo}` -> no
`${_MY234}`
`${$MY}`

glours: we can't do something generic like this, because it could reference either boolean, integer values or more complex structs like ports or range of IPs which could be defined in multiple ways... and I'm not sure we want to redeclare all kind of type as we know compose-go will do the check when parsing the compose model

jhrotko: maybe I am understanding this wrong but my understanding is when you write `${VAR}`, the schema will not interpret this value. This will be done by compose-go when loading the file (right?) So, in my point of view, if you want the schema to validate the current value `${VAR}` - which is a simple string with a syntax defined in compose-go

ndeloof: variable syntax is way more complicated. Can be `${VAR:-${SUBVAR?err}}`
Also `delay: 12${VARIABLE}s` is a valid value

ndeloof: > if you want the schema to validate the current value ${VAR} is valid

right, I don't want this to happen. I expect schema to validate the yaml tree structure, so I can safely assume I can access `services.xx.mac_address`. But interpolating value and validating a valid mac address should be implemented in compose-go, not by schema

jhrotko: I see, it can be more complex than just a simple `$VAR`
Screenshot 2024-07-02 at 12 40 07

jhrotko (approved):

PR #654, opened 2024-07-04
keep empty environment variables as those must be UNSET in container

ndeloof:
environment `FOO` with no value nor equal sign means "propagate value from user's environment, or unset variable"
So, as we resolve environment, empty value with no matching entry in user's environment MUST not result into removing entry from service definition.
Build args do not follow this rule. `--build-arg FOO` will propagate user's environment value, but if not set build arg will keep using value set by Dockerfile

fixes https://github.com/docker/compose/issues/11962
supersede https://github.com/compose-spec/compose-go/pull/634
see https://github.com/docker/compose/pull/11965

### DIFF
--- a/loader/normalize.go
+++ b/loader/normalize.go
@@ -52,14 +52,14 @@ func Normalize(dict map[string]any, env types.Mapping) (map[string]any, error) {
 				}
 
 				if a, ok := build["args"]; ok {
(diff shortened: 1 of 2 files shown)


glours (approved):
👍

jhrotko (approved):
LGTM

PR #655, opened 2024-07-04
load secrets/configs from environment while parsing compose model

ndeloof:
As we load compose model with environment, resolve secrets and config values based on declared environment value. 
This allows a nested model loading (using `include`) to resolve such values based on configured `env_file`, not just top-level environment.
when marshaling model to yaml/json resolved content is hidden if it was set by environment, to keep definition consistency

closes https://github.com/docker/compose/issues/11952

### DIFF
--- a/loader/environment.go
+++ b/loader/environment.go
@@ -22,9 +22,14 @@ import (
 	"github.com/compose-spec/compose-go/v2/types"
 )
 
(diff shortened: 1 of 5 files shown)


glours (approved):
LGTM

jhrotko (approved):
LGTM

PR #661, opened 2024-07-10
Resolve symbolic links when loading develop.watch.path

jhrotko:
Resolve symbolic links in `develop.watch.path` when loading the project

fixes: https://github.com/docker/compose/issues/11849


ndeloof:
Can't we just rely on https://pkg.go.dev/path/filepath#EvalSymlinks here ?

jhrotko:
> Can't we just rely on https://pkg.go.dev/path/filepath#EvalSymlinks here ?
@ndeloof, the thing is, the directory is a symbolic link and not the file itself. I tried using `EvalSymlinks` but it was not enough for this case.

> I also wonder we should do this for all path made absolute, not just watch where this had some side effect, wdyt?
I was not sure myself. I think the advantage of adding this change here, you do not need to deal with any transformations going further to compose. I can see both solutions working fine in this case

ndeloof (approved):

PR #665, opened 2024-07-22
resolveRelativePaths can run with non-canonical model

ndeloof:
closes https://github.com/docker/compose/issues/11994


glours (approved):
LGTM

PR #666, opened 2024-07-23
Revert "run interpolation after merge, but for required attributes"

idsulik:
This reverts commit 65600cee45d45771a1faa6ddaf87b23ca4d2400c.

The commit led to this issue https://github.com/docker/compose/issues/12001, a quick fix is to revert the commit and release it.

### DIFF
--- a/cli/options_test.go
+++ b/cli/options_test.go
@@ -253,7 +253,7 @@ func TestProjectWithDotEnv(t *testing.T) {
 		"compose-with-variables.yaml",
 	}, WithName("my_project"), WithEnvFiles(), WithDotEnv)
 	assert.NilError(t, err)
-	p, err := opts.LoadProject(context.TODO())
+	p, err := ProjectFromOptions(context.TODO(), opts)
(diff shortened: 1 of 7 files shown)


ndeloof:
We only do such revert if there's no reasonable fix to offer.
For those impacted it is easy to just keep running older docker compose release

idsulik:
> We only do such revert if there's no reasonable fix to offer. For those impacted it is easy to just keep running older docker compose release

Okay, I got it. It was just a quick fix.  I think it'll be better to revert and release if we don't be able to fix it asap, because more and more users download the latest version and a lot of them will be forced to downgrade or spent time figuring out what went wrong

ndeloof:
Sure. A fix is proposed by @jhrotko here : https://github.com/compose-spec/compose-go/pull/667

idsulik:
Great!

jhrotko:
there are more issues than in #667

ndeloof (approved):

jhrotko (approved):

PR #667, opened 2024-07-23
Escape double dollar during include process

jhrotko:


loader/include.go
```diff
@@ -212,3 +217,24 @@ func importResource(source map[string]any, target map[string]any, key string) er
...
+		return strings.ReplaceAll(v, "$", "$$")
+	case []any:
+		for i, el := range v {
+			v[i] = escape(el)
+		}
+		return v
+	case map[string]any:
+		for k, el := range v {
+			v[k] = escape(el)
+		}
```
ndeloof, line 237: `return v`

PR #669, opened 2024-07-30
feat(compose-spec): Add content property for secrets in compose-spec.json

idsulik:
Fixes https://github.com/docker/compose/issues/12033

Add `content` to `secret` property, because in case of environment value, we add "content" property but it wasn't defined in the spec

### DIFF
--- a/loader/environment.go
+++ b/loader/environment.go
@@ -79,7 +79,7 @@ func resolveSecretsEnvironment(dict map[string]any, environment types.Mapping) {
 			continue
 		}
 		if found, ok := environment[env]; ok {
-			secret["content"] = found
+			secret[types.SecretConfigXValue] = found
 		}
 		secrets[name] = secret
 	}
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -774,6 +774,11 @@ func secretConfigDecoderHook(from, to reflect.Type, data interface{}) (interface
 				if val, ok := ext[types.SecretConfigXValue].(string); ok {
 					// Return a map with the Content field populated
 					v["Content"] = val
+					delete(ext, types.SecretConfigXValue)
+
+					if len(ext) == 0 {
+						delete(v, "#extensions")
+					}
 				}
 			}
 		}
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -738,7 +738,9 @@ func Transform(source interface{}, target interface{}) error {
 		DecodeHook: mapstructure.ComposeDecodeHookFunc(
 			nameServices,
 			decoderHook,
-			cast),
+			cast,
+			secretConfigDecoderHook,
+		),
 		Result:   target,
 		TagName:  "yaml",
 		Metadata: &data,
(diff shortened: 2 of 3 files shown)


ndeloof:
> we add "content" property but it wasn't defined in the spec

This is by design: we don't want sensible data to be exposed as plain text in a compose file

ndeloof (changes requested):
This introduces a major risk for users to put sensible data in a compose file and publish by mistake, we can't let them shoot into their own feet

idsulik:
@ndeloof good point, thank you! pushed fix.  Added validation to prevent from specifying "content" for the secrets property

validation/validation.go
```diff
@@ -61,6 +61,19 @@ func check(value any, p tree.Path) error {
...
 }
 
+func checkSecrets(value any, p tree.Path) error {
+	if err := checkFileObject("file", "environment")(value, p); err != nil {
+		return err
+	}
+
+	v := value.(map[string]any)
+	if _, ok := v["content"]; ok {
+		return fmt.Errorf("%s: content is not allowed for secrets", p)
```
ndeloof, line 71: It doesn't make sense to have this made valid by schema but rejected by parser.

idsulik: so I see only one way - to revert your changes where you added "content" to the "secrets" property

idsulik: @ndeloof if you have any other ideas, please share

ndeloof:
`SecretConfig` already has a custom `MarshalJSON` so that `content` is excluded from json output, I wonder why gojsonschema does not rely on this to validate the model.

schema/schema.go
```diff
@@ -72,6 +73,19 @@ func Validate(config map[string]interface{}) error {
 	return nil
 }
 
+// removeSecretsContentProperty removes the content property from secrets
+// we add the content key here loader/environment.go:66
+func removeSecretsContentProperty(config map[string]interface{}) map[string]interface{} {
```
idsulik, line 78: @ndeloof here is an equivalent ) at least tmp solution

ndeloof: but we loose value doing so

idsulik: we can copy for validation

idsulik: pushed changes. now it clones the map and validate

ndeloof (commented):
AFAICT this won't catch a compose file using invalid attribute `content` to set secret, while value will be actually loaded.

idsulik:
> Suggestion: to avoid letting `content` be used in compose.yaml and miss it during validation, resolved value could be stored in map with key `#value` (with a `#`) which makes it an invalid key in yaml (until user set key with quote, but this can't be by accident then).

@ndeloof , if you mean this changes:
```
# environment.go
		if found, ok := environment[env]; ok {
			secret["#value"] = found
		}
```
then it won'r work:

> secrets.one-secret-token Additional property #value is not allowed

ndeloof:
oh indeed.

idsulik:
> A possible workaround is to use a fake extension key `x-#value` that would be valid regarding the spec

I'm not sure I get the idea if I rename 
```
if found, ok := environment[env]; ok {
	secret["content"] = found
}
```
into
```
if found, ok := environment[env]; ok {
	secret["x-#value"] = found
}
```
Then how do we map this key into the Secret struct's `content` field?

ndeloof:
> Then how do we map this key into the Secret struct's content field?

need to add a `DecodeMapstructure` func to `FileObjectConfig` so it detects this custom key is set and use it as value

idsulik:
@ndeloof pushed changes.  
p.s. extracted the key name into constant for consistency

glours (approved):
LGTM
@idsulik can you sign-off your commits please?

idsulik:
@glours done

PR #670, opened 2024-07-31
Allow empty lists

lionello:
This happens when some users are not very familiar with the Compose file syntax: during testing they'd comment out the environment variables, but not the actual `environment:` key, and the Compose file would fail to load:
```
services:
  service1:
    build: .
    environment:
      #NODE_ENV: production
```

This change adds `"null"` as a valid type for many of the lists in the JSON schema.


lionello:
If this is the right approach I'm OK to add `"oneOf": [{"type": "null"}, {"type": "array": …}]` to the remaining ones.

schema/compose-spec.json
```diff
@@ -103,12 +108,12 @@
                 "context": {"type": "string"},
                 "dockerfile": {"type": "string"},
                 "dockerfile_inline": {"type": "string"},
-                "entitlements": {"type": "array", "items": {"type": "string"}},
+                "entitlements": {"$ref": "#/definitions/list_of_strings"},
```
lionello, line 111: Some of these could possibly be `list_of_unique_strings` but that would technically be a breaking change.

jhrotko:
they should get a warning from the IDE (if an extension is installed) or they get an error when running `up` for instance. The current behaviour seems is expected if `environment` or any other compose key is empty.

lionello:
> they should get a warning from the IDE (if an extension is installed) or they get an error when running `up` for instance. The current behaviour is expected if `environment` or any other compose key is empty.

That's right, you do get an error when you do `up`, but IMHO handling `null` is a better dev experience. Because arrays can be `nil` (and a `nil` array has length 0), the code already handles this.

ndeloof:
I agree there's no reason we consider an `null` list as invalid while an empty one if fine

PR #673, opened 2024-08-10
feat(volume): Add device with colon support

idsulik:
Closes #https://github.com/docker/compose/issues/12052

Added device with colon support, we used to support it before this commit https://github.com/compose-spec/compose-go/commit/072aa6418dcd93b4e31d0235df9dad5a6c11e63e

Example:
```
services:
  app:
    image: alpine
    command: ["sh", "-c", "while true; do echo 'Hello, World!'; sleep 3; done"]
    devices:
      - /dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0-port0:/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0-port0:ro
      - /dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.3:1.0-port0:/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.3:1.0-port0
```

### DIFF
--- a/format/volume.go
+++ b/format/volume.go
@@ -41,6 +41,29 @@ func ParseVolume(spec string) (types.ServiceVolumeConfig, error) {
 		return volume, nil
 	}
 
+	if strings.Count(spec, "/dev") == 2 {
+		parts := strings.Split(spec, ":/dev")
+		source := parts[0]
+		target := "/dev" + parts[1]
+
+		if err := populateFieldFromBuffer(':', []rune(source), &volume); err != nil {
+			populateType(&volume)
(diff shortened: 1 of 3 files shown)


ndeloof:
I'm not a fan we introduce a special handling for `/dev` prefix as `docker run` does not support this either.
IMHO better introduce explicit support for "long syntax" in compose specification when setting a device mount, so user can disambiguate such a confusing path

idsulik:
I was going to add long syntax support myself, but it's already been done here https://github.com/compose-spec/compose-go/pull/676#pullrequestreview-2245052117

PR #676, opened 2024-08-19
long syntax for device mapping

ndeloof:
introduce long syntax for device mapping, to help disambiguate path with colons

closes https://github.com/docker/compose/issues/12052

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -3329,3 +3329,51 @@ secrets:
 			Content:     "Shadoks",
 		}})
 }
+
+func TestLoadDeviceMapping(t *testing.T) {
+	config, err := loadYAML(`
+name: load-device-mapping
+services:
+  test:
+    devices:
+      - /dev/source:/dev/target:permissions
+      - /dev/single
+`)
+	assert.NilError(t, err)
+	assert.DeepEqual(t, config.Services["test"].Devices, []types.DeviceMapping{
+		{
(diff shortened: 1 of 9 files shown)


ndeloof:
cc @idsulik

types/types.go
```diff
@@ -301,6 +301,12 @@ type BlkioConfig struct {
 	Extensions Extensions `yaml:"#extensions,inline,omitempty" json:"-"`
 }
 
+type DeviceMapping struct {
+	Source      string `yaml:"source,omitempty" json:"source,omitempty"`
+	Target      string `yaml:"target,omitempty" json:"target,omitempty"`
+	Permissions string `yaml:"permissions,omitempty" json:"permissions,omitempty"`
+}
```
jhrotko, line 308: Missing Extension? From `compse-spec` you defined the pattern for Extensions

loader/loader_test.go
```diff
@@ -3329,3 +3329,27 @@ secrets:
 			Content:     "Shadoks",
 		}})
 }
+
+func TestLoadDeviceMapping(t *testing.T) {
+	config, err := loadYAML(`
+name: load-device-mapping
+services:
+  test:
+    devices:
```
jhrotko, line 3338: Could you also introduce the test case for the object/map case (with an extension)?

jhrotko (approved):

PR #678, opened 2024-09-02
Support multiple values for extra_hosts

ndeloof:
This PR adds support for multiple-value set by `extra_hosts`, either with long syntax _or_ single line syntax, using a coma separator between IPs in value. Unicity doesn't apply to extra_hosts as this `docker run` flag actually supports multiple values per key being set, so it was wrong for compose to consider an override.

closes https://github.com/docker/compose/issues/12088

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -3377,3 +3377,35 @@ services:
 		},
 	})
 }
+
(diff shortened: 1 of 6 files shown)


jhrotko (approved):
LGTM

tate-button:
big thanks @ndeloof it's much appreciated

Issue #668, opened 2024-07-30
Panic crash if services are not indented correctly with extended compose files.


DocFX:
Hi, I just had a weird crash (took me some time to figure it out ^^).

Stack trace:

```
PS F:\Important\xxx\git\xxx> docker compose --env-file "./environment-files/.env" --env-file "./environment-files/.env.prod-zone-1-server-2" --env-file "./environment-files/.env.prod-zone-1-server-2.local" -f "./.devcontainer/docker-compose-files/xxx-containers-compose-prod-zone-1-server-2.yml" config
panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 1 [running]:
github.com/compose-spec/compose-go/v2/loader.getExtendsBaseFromFile({0x2ba8da8, 0xc000538870}, {0xc0002f6080, 0x1c}, {0xc0006877e8, 0x14}, {0xc0003b8b00, 0x7f}, {0xc0005347b0, 0x28}, ...)
        github.com/compose-spec/compose-go/v2@v2.1.5/loader/extends.go:166 +0x9bc
github.com/compose-spec/compose-go/v2/loader.applyServiceExtends({0x2ba8da8?, 0xc000538870?}, {0xc0002f6080, 0x1c}, 0xc000539aa0, 0xc000536240, 0x4?, {0xc0004968a8, 0x1, 0x1})
        github.com/compose-spec/compose-go/v2@v2.1.5/loader/extends.go:86 +0x44f
github.com/compose-spec/compose-go/v2/loader.ApplyExtends({0x2ba8da8, 0xc000538870}, 0xc0003f1260?, 0x2957158?, 0xc000496768?, {0xc0004968a8, 0x1, 0x1})
        github.com/compose-spec/compose-go/v2@v2.1.5/loader/extends.go:40 +0x178
github.com/compose-spec/compose-go/v2/loader.loadYamlFile.func1({0x243dce0?, 0xc000538a50?}, {0xc0004968a8, 0x1, 0x1})
        github.com/compose-spec/compose-go/v2@v2.1.5/loader/loader.go:428 +0x16d
github.com/compose-spec/compose-go/v2/loader.loadYamlFile({0x2ba8de0, 0xc000444780}, {{0xc0003b8b00, 0x7f}, {0xc00050c700, 0x6f5, 0x6f6}, 0x0}, 0xc000536240, {0xc000057360, ...}, ...)
[...]

PR #679, opened 2024-09-05
prevent panic accesing services in extended file

ndeloof:
closes https://github.com/compose-spec/compose-go/issues/668


glours (approved):
LGTM

Issue #663, opened 2024-07-19
Dashes in variable names <=> `environment:` block


simonhammes:
Dashes can be used in variable names (https://github.com/compose-spec/compose-go/pull/336).

However, I'm unable to use a variable that contains dashes in an `environment:` block since `compose` interprets the dash as a separator between the variable name and its default value:

```yml
environment:
  VAR-WITH-DASHES: ${VAR-WITH-DASHES}
```

Is there any way to escape the dash or is this a bug/unimplemented functionality?

ndeloof:
`-` indeed is delimiter for default value, which makes #336 actually irrelevant 😓 
I guess we should revert, as this issue demonstrate this "feature" was incomplete and can't be used anyway
cc @jhrotko @glours

KasonBraley:
@glours Can this bug fix be released in a patch so [compose](https://github.com/docker/compose) can release a version with the fix?

ndeloof:
no worries, this will be part of the next compose release coming soon.
Cheers

PR #680, opened 2024-09-05
Revert "accept dashes in variable names"

ndeloof:
This reverts commit be170befda31a336a94d248ac1221fceeac834c8.

we **can't** support dashes in variables as this conflict with the [interpolation syntax](https://github.com/compose-spec/compose-spec/blob/main/12-interpolation.md) for default values

closes https://github.com/compose-spec/compose-go/issues/663


glours (approved):
LGTM

PR #684, opened 2024-09-24
Update compose-spec.json: remove invalid fields from array

AnaMManolache:
The 'additionalProperties' and 'patternProperties' are being declared for an array type, which is invalid. They are also declared for the items of the array, so they can be removed. Their presence is causing issues converting this jsonschema to its cuelang representation.

Fixes https://github.com/compose-spec/compose-spec/issues/534

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -500,9 +500,7 @@
             },
             "additionalProperties": false,
             "patternProperties": {"^x-": {}}
-          },
-          "additionalProperties": false,
-          "patternProperties": {"^x-": {}}
+          }
         }
       }
     },
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -502,7 +502,9 @@
             "patternProperties": {"^x-": {}}
           }
         }
-      }
+      },
+      "additionalProperties": false,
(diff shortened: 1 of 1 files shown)


ndeloof (approved):

ndeloof (changes requested):
additionalProperties is indeed wrongly declared on `watch` type, while it should apply to the parent `development` element. Should not be removed but moved to parent type

schema/compose-spec.json
```diff
@@ -500,9 +500,7 @@
             },
             "additionalProperties": false,
             "patternProperties": {"^x-": {}}
-          },
-          "additionalProperties": false,
-          "patternProperties": {"^x-": {}}
```
glours, line 505: Should be moved to the `development` level as this section is missing the capability to add extension attributes `x-`

glours (changes requested):

schema/compose-spec.json
AnaMManolache: Added them back on development

ndeloof (approved):

glours (approved):
LGTM, thanks for your contribution 🙏

PR #687, opened 2024-09-30
allow use of x-* service names as depends_on keys

ndeloof:
fix https://github.com/docker/compose/issues/12171

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -3409,3 +3409,24 @@ services:
 		"myhost": []string{"0.0.0.1", "0.0.0.2"},
 	})
 }
+
+func TestLoadDependsOnX(t *testing.T) {
+	p, err := loadYAML(`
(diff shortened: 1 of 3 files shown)


loader/loader_test.go
```diff
@@ -3409,3 +3409,24 @@ services:
...
+
+func TestLoadDependsOnX(t *testing.T) {
+	p, err := loadYAML(`
+name: load-depends-on-x
+services:
+  test:
+    image: test
+    depends_on:
+      - x-foo
+  foo:
```
glours, line 3421: It should be `x-foo`, no?
```suggestion
  x-foo:
```

ndeloof: indeed, not actually used by the test

glours (approved):
LGTM

jhrotko (approved):
LGTM

PR #688, opened 2024-09-30
fix subpath in ServiceVolumeConfig#String

ndeloof:
quick fix as subpath option is malformed

To be debated: as this option is actually not supported in the string format, why not just exclude it 
_or_ why don't we also include createHostPath then ? #inconsistency

### DIFF
--- a/format/volume.go
+++ b/format/volume.go
@@ -95,7 +95,7 @@ func populateFieldFromBuffer(char rune, buffer []rune, volume *types.ServiceVolu
 			if isBindOption(option) {
 				setBindOption(volume, option)
 			}
-			// ignore unknown options
+			// ignore unknown options FIXME why not report an error here?
 		}
 	}
 	return nil
--- a/types/types.go
+++ b/types/types.go
@@ -552,9 +552,6 @@ func (s ServiceVolumeConfig) String() string {
 	if s.Volume != nil && s.Volume.NoCopy {
 		options = append(options, "nocopy")
 	}
-	if s.Volume != nil && s.Volume.Subpath != "" {
-		options = append(options, s.Volume.Subpath)
-	}
 	return fmt.Sprintf("%s:%s:%s", s.Source, s.Target, strings.Join(options, ","))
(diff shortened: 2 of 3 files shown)


ndeloof:
As docker run -v rejects unknown bind options, maybe better just remove `subpath` from String, WDYT ?

```
$ docker run -v .:/foo:ro,subpath=test alpine true
docker: Error response from daemon: invalid mode: ro,subpath=test.
```

types/types.go
```diff
@@ -552,9 +552,6 @@ func (s ServiceVolumeConfig) String() string {
 	if s.Volume != nil && s.Volume.NoCopy {
 		options = append(options, "nocopy")
 	}
-	if s.Volume != nil && s.Volume.Subpath != "" {
-		options = append(options, s.Volume.Subpath)
-	}
```
glours, line 557: Just to be sure, we have a warning in Compose to let the users know it won't be used?
What adding the warning here for all the other implementations?

ndeloof: nope, Docker Compose **HAS** to use the mount API when `SubPath` or `CreateHostPath` is set

glours (approved):

format/volume.go
```diff
@@ -95,7 +95,7 @@ func populateFieldFromBuffer(char rune, buffer []rune, volume *types.ServiceVolu
 			if isBindOption(option) {
 				setBindOption(volume, option)
 			}
-			// ignore unknown options
+			// ignore unknown options FIXME why not report an error here?
```
jhrotko, line 98: I would vote to add a message here since the engine error message becomes cryptic to the user since these modes are set by compose so the engine can do its thing

jhrotko (approved):

format/volume.go
ndeloof: The engine has it's own parser for the bind string, which reject unknown options.
Docker Compose just must not send a bind string to engine when it can't translate into, but `String` can't return an error

PR #690, opened 2024-10-01
introduce ability to use custom env_file format

ndeloof:
This allows to register additional env_file format parser, so one can plug https://github.com/docker/cli/blob/master/opts/envfile.go as an alternate parser, to support reading "raw" env_file just like `docker run --env-file` does.

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -616,22 +616,11 @@ func (p Project) WithServicesEnvironmentResolved(discardEnvFiles bool) (*Project
 		}
 
 		for _, envFile := range service.EnvFiles {
-			if _, err := os.Stat(envFile.Path); os.IsNotExist(err) {
-				if envFile.Required {
-					return nil, fmt.Errorf("env file %s not found: %w", envFile.Path, err)
-				}
-				continue
-			}
-			b, err := os.ReadFile(envFile.Path)
+			vars, err := loadEnvFile(envFile, resolve)
 			if err != nil {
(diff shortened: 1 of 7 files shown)


types/project.go
```diff
@@ -644,6 +633,31 @@ func (p Project) WithServicesEnvironmentResolved(discardEnvFiles bool) (*Project
...
+	defer file.Close() //nolint:errcheck
+
+	var fileVars map[string]string
+	if envFile.Format != "" {
+		fileVars, err = dotenv.ParseWithFormat(file, envFile.Path, resolve, envFile.Format)
+	} else {
+		fileVars, err = dotenv.ParseWithLookup(file, resolve)
+	}
+	if err != nil {
+		return nil, nil
```
jhrotko, line 656:
```suggestion
		return nil, err
```

jhrotko (changes requested):
LGTM! I would also add/change a test in `loader` to also cover this case

types/project.go
ndeloof: indeed 😓

ndeloof:
@jhrotko a test in `loader` would make sense if we had an actual alternate format implementation hosted in this repo, but here we just define an extension point. raw loader test is [here](https://github.com/docker/compose/pull/12179/files#diff-f16f866ca631acf1a7182a4354c0e9bd3c0c2ebcb56c5a8754a703a5585037f6)

jhrotko (approved):
LGTM!

glours (approved):
LGTM

PR #691, opened 2024-10-02
add (missing) deploy.resources.devices.options

ndeloof:
defined by the spec but missing in go struct for some legacy reasons

### DIFF
--- a/types/device.go
+++ b/types/device.go
@@ -27,6 +27,7 @@ type DeviceRequest struct {
 	Driver       string      `yaml:"driver,omitempty" json:"driver,omitempty"`
 	Count        DeviceCount `yaml:"count,omitempty" json:"count,omitempty"`
 	IDs          []string    `yaml:"device_ids,omitempty" json:"device_ids,omitempty"`
+	Options      Mapping     `yaml:"options,omitempty" json:"options,omitempty"`
 }
 
 type DeviceCount int64
@@ -91,6 +92,13 @@ func (d *DeviceRequest) DecodeMapstructure(value interface{}) error {
 		}
 		d.IDs = ids
 		d.Count = DeviceCount(len(ids))
+	}
+
+	d.Options = Mapping{}
+	if options, ok := v["options"].(map[string]any); ok {
+		for k, v := range options {
+			d.Options[k] = fmt.Sprint(v)
+		}
 	}
 	return nil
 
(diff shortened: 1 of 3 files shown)


types/device.go
```diff
@@ -27,6 +27,7 @@ type DeviceRequest struct {
 	Driver       string      `yaml:"driver,omitempty" json:"driver,omitempty"`
 	Count        DeviceCount `yaml:"count,omitempty" json:"count,omitempty"`
 	IDs          []string    `yaml:"device_ids,omitempty" json:"device_ids,omitempty"`
+	Options      Mapping     `yaml:"options,omitempty" json:"options,omitempty"`
```
jhrotko, line 30: We are not doing anything with this yet. Should we wait for the upcoming work to add this?

ndeloof: we are not because this was missing in compose-go, but [moby API](https://docs.docker.com/reference/api/engine/version/v1.45/#tag/Container/operation/ContainerCreate) has support for options (sorry, can't post a direct link to DeviceRequest nested struct)
> Driver-specific options, specified as a key/value pairs. These options are passed directly to the driver.

jhrotko: Apologies, re-reading my comment I realise I did not express myself correctly. I meant, in compose-go you defined the `Options` but not doing any actual implementation with it. All I am saying is should we save this change for later for when we augment the options section or should we already implement the moby api call in this PR?

ndeloof: compose-go never includes implementation; this is docker/compose role to do so

ndeloof: indeed, thanks

ndeloof: fixed, +added a test case to cover loading DeviceRequest

types/device.go
```diff
@@ -92,6 +93,13 @@ func (d *DeviceRequest) DecodeMapstructure(value interface{}) error {
 		d.IDs = ids
 		d.Count = DeviceCount(len(ids))
 	}
+
+	if options, ok := v["options"].(map[string]any); ok {
+		d.Options = Mapping{}
```
glours, line 98: Can't we do this even if there isn't any option defined to avoid nil issue on the client side?

glours (approved):
LGTM

types/device.go
jhrotko: @ndeloof @glours thanks!

jhrotko (approved):
LGTM

PR #692, opened 2024-10-04
prefer canonical transform vs DecodeMapStructure

ndeloof:
Avoid use of DecodeMapstructure when we can just make yaml tree canonical


ndeloof:
I haven't removed DeviceCount#DecodeMapstructure which handle this case

glours:
I just see that, and the `all` is converted to `-1` so it's fine for me 👍

glours (approved):
LGTM

ndeloof:
to be debated such moby-specific tweaks should live in this repository, or must move to docker specific implementation as docker/compose

PR #694, opened 2024-10-07
introduce service hooks

ndeloof:
Introduce Service lifecycle hooks to run commands on `post_start` or `pre_stop` inside service container

### DIFF
--- a/types/hooks.go
+++ b/types/hooks.go
@@ -1,0 +1,26 @@
+/*
+   Copyright 2020 The Compose Specification Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


types/hooks.go
```diff
@@ -0,0 +1,26 @@
...
+package types
+
+// ServiceHook is a command to exec inside container by some lifecycle events
+type ServiceHook struct {
+	Command     ShellCommand      `yaml:"command,omitempty" json:"command"`
+	User        string            `yaml:"user,omitempty" json:"user,omitempty"`
+	Privileged  bool              `yaml:"privileged,omitempty" json:"privileged,omitempty"`
+	WorkingDir  string            `yaml:"working_dir,omitempty" json:"working_dir,omitempty"`
+	Environment MappingWithEquals `yaml:"environment,omitempty" json:"environment,omitempty"`
+}
```
glours, line 26: aren't we missing the extensions here?
```suggestion
	Environment MappingWithEquals `yaml:"environment,omitempty" json:"environment,omitempty"`
	
	Extensions Extensions `yaml:"#extensions,inline,omitempty" json:"-"`
}
```

glours (approved):
Sounds good to me, I'm just wondering if we aren't just missing the `Extensions` attribut in the `ServiceHook` struct

types/hooks.go
ndeloof: fixed by a59035ad2bf454979d724b482c403efde4d8bee9

laurazard:
Nice!!! 🎉

PR #696, opened 2024-10-11
seiralise NanoCPUs as string

ndeloof:
value without a decimal part will be rendered as integer, so need to also accept integer

fixes https://github.com/docker/compose/issues/11903

this PR only adds `case int:` but I used this opportunity to extract type in a dedicated go file


glours (approved):
LGTM

PR #697, opened 2024-10-14
normalize volume.target to drop trailing slash

ndeloof:


jhrotko (approved):
So, we moved the clean path to compose-go and cleaned up from compose in order to avoid inconsistensies?

ndeloof:
in docker/compose we ran `volume.Target = path.Clean(volume.Target)` within a func
1. this was buggy as caller value was left unchanged (would need to pass a pointer)
2. this is unsafe we "fix" the model inside some unrelated subfunction
with this change, we get volume target cleaned after model is loaded, so we can safely compare values.

PR #698, opened 2024-10-15
add support for bind mount recursive

ndeloof:
add support for bind mount recursive settings (https://docs.docker.com/engine/storage/bind-mounts/#recursive-mounts)
see https://github.com/docker/compose/issues/11792

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -419,6 +419,7 @@
                     "properties": {
                       "propagation": {"type": "string"},
                       "create_host_path": {"type": ["boolean", "string"]},
+                      "recursive": {"type": "string", "enum": ["enabled", "disabled", "writable", "readonly"]},
                       "selinux": {"type": "string", "enum": ["z", "Z"]}
                     },
                     "additionalProperties": false,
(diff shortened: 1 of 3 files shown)


schema/compose-spec.json
```diff
@@ -419,7 +419,8 @@
                     "properties": {
                       "propagation": {"type": "string"},
                       "create_host_path": {"type": ["boolean", "string"]},
-                      "selinux": {"type": "string", "enum": ["z", "Z"]}
+                      "recursive": {"type": "string", "enum": ["z", "Z"]},
+                      "selinux": {"type": "string", "enum": ["enabled", "disabled", "writable", "readonly"]}
```
glours, line 423: Inversion?
```suggestion
                      "selinux": {"type": "string", "enum": ["z", "Z"]},
                      "recursive": {"type": "string", "enum": ["enabled", "disabled", "writable", "readonly"]}
```

jhrotko (approved):
LGTM

schema/compose-spec.json
ndeloof: indeed. I'm trying to reduce my caffeine addiction, but this seems to come with side effects

glours (approved):
LGTM

PR #700, opened 2024-10-17
introduce OmitEmpty in yaml processing pipeline

ndeloof:
Introduce a generic mechanism to drop attributes with an empty value. A typical scenario is: attribute is set by an interpolation variable, which resolves to empty string (maybe variable is unset? Then other already got a warning). Some attribute are just irrelevant with an empty value, so need to be ignored. Better just remove those from model

This first iteration drops empty `dns` as requested on https://github.com/docker/compose/issues/11690
This relies on the path pattern syntax used in other yaml processing stages, so we can easily add more "omit" rules.

closes https://github.com/docker/compose/issues/11690

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -516,6 +516,8 @@ func loadYamlFile(ctx context.Context, file types.ConfigFile, opts *Options, wor
 			return err
 		}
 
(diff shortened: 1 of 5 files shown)


glours (approved):
LGTM

PR #701, opened 2024-10-22
load configuration file with right ResourceLoader

glours:
Detect and use the right ResourceLoader to load a config file and add it `ConfigDetails` struct.

`LocalResourceLoader` is always used as the last loader and should failed if it's not able to parse the configuration file.

This PR replace #592

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	"github.com/compose-spec/compose-go/v2/consts"
+	"github.com/compose-spec/compose-go/v2/errdefs"
 	interp "github.com/compose-spec/compose-go/v2/interpolation"
 	"github.com/compose-spec/compose-go/v2/override"
 	"github.com/compose-spec/compose-go/v2/paths"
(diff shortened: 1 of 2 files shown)


loader/loader.go
```diff
@@ -300,6 +301,51 @@ func parseYAML(decoder *yaml.Decoder) (map[string]interface{}, PostProcessor, er
...
+	}
+
+	for _, op := range options {
+		op(opts)
+	}
+	opts.ResourceLoaders = append(opts.ResourceLoaders, localResourceLoader{})
+
+	for i, p := range configFiles {
+		for _, loader := range opts.ResourceLoaders {
+			_, isLocalResourceLoader := loader.(localResourceLoader)
```
ndeloof, line 323: if local loader reject `Accept` it's highly probable it will also fail on `Load`, so why not just let it reject path and get this function return an error if no loader accepted the path ?

ndeloof (approved):
LGTM, some minor comment

loader/loader.go
glours: Yes but you won't have the reason why, what I could do is changing the `Accept` signature to return an error in addition of the boolean, WDYT?

ndeloof: the other option could be for `localResource Loader` to always accept path. The `Load` can return file not found error

glours: Ok I'll adapt the `Accept` function of resource loader

ndeloof (approved):

PR #702, opened 2024-10-24
detect project dir is a symlink and warn user

ndeloof:
when ran from a symlink, while we find compose.yaml we run with a distinct project_dir and project_name. This might be user intent but also can cause unexpected behavior. This PR adds a sanity-check for such circumstances to warn user

closes https://github.com/docker/compose/issues/12026


glours (approved):
LGTM

PR #703, opened 2024-10-26
fix(reset): Add cycle detector in reset.go

idsulik:
Issue: https://github.com/docker/compose/issues/12235
I added the cycle detector to `reset.go`.  
Example yaml to reproduce the issue:
```
x-healthcheck: &healthcheck
  healthcheck:
    test: [ "CMD", "wget", "--quiet", "--spider", "--timeout=1", "http://localhost:9000/health/live" ]
    start_period: 20s
    start_interval: 1s
    interval: 30s
    timeout: 2s
    retries: 5


  egress-service:
    <<: *healthcheck
```

PR #704, opened 2024-10-29
fix(reset): Add cycle detection fix for services that are prefixed with the name of a preceding service

idsulik:
Fixed the cycle detector https://github.com/compose-spec/compose-go/pull/703  

It detects cycle for services that are prefixed with the name of a preceding service
https://github.com/docker/compose/issues/12235#issuecomment-2445030390


glours (approved):
LGTM, thanks @idsulik for the quick fix and for adding the test 🙏

PR #706, opened 2024-10-31
fix(reset): Add cycle detection fix

idsulik:
Here https://github.com/docker/compose/issues/12247#issuecomment-2448134930 the user found one more case when it fires while it shouldn't.  
1. Decided to move the detector inside the alias condition, because that's the only place where the cycle can happen.
2. Added one more test


idsulik:
@glours hi! Check please, I think it's pretty important

glours (approved):

PR #709, opened 2024-11-06
fix(reset): Improve cycle detector

idsulik:
Improved the cycle detector and added one more test case.  

https://github.com/docker/compose/issues/12247#issuecomment-2459965906

### DIFF
--- a/loader/reset.go
+++ b/loader/reset.go
@@ -28,11 +28,12 @@ import (
 type ResetProcessor struct {
 	target       interface{}
 	paths        []tree.Path
-	visitedNodes map[*yaml.Node]string
+	visitedNodes map[*yaml.Node][]string
 }
 
 // UnmarshalYAML implement yaml.Unmarshaler
 func (p *ResetProcessor) UnmarshalYAML(value *yaml.Node) error {
+	p.visitedNodes = make(map[*yaml.Node][]string)
 	resolved, err := p.resolveReset(value, tree.NewPath())
 	p.visitedNodes = nil
 	if err != nil {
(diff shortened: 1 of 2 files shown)


loader/reset.go
```diff
@@ -144,20 +145,46 @@ func (p *ResetProcessor) applyNullOverrides(target any, path tree.Path) error {
...
+	}
+
+	p.visitedNodes[node] = append(paths, pathStr)
 	return nil
 }
+
+// areInDifferentServices checks if two paths are in different service definitions
+func areInDifferentServices(path1, path2 string) bool {
+	// Split paths into components
+	parts1 := strings.Split(path1, ".")
```
ndeloof, line 178: nit: could use path.Parts() if you don't convert it to a string

idsulik: the function accepts strings and the field also has a list of strings `visitedNodes map[*yaml.Node][]string`, do you suggest to use path everywhere? or to convert the strings inside the func into paths and use `Parts()`? it looks like overhead, isn't it?

ndeloof: some overhead indeed, but my point is that `Path` abstraction allows to ignore "implementation details" like the dot notation and `[]` or `*` patterns. I had in mind at some point we might want to use an alternate implementation vs `string` as we actually have to split this string many times during parsing.
From an encapsulation point of view, I'd prefer we add more utility func to `Path` to support required submatching. 
ANYWAY I'm fine we merge this PR as a quick fix for this issue, and revisit this later

ndeloof (approved):

glours (approved):
Let's do a follow up PR for the improvements

loader/reset.go
idsulik: @ndeloof agree, that sounds reasonable.

PR #711, opened 2024-11-12
fix(variables): Add number suffix support to the variable extractor

idsulik:
Fixes https://github.com/docker/compose/issues/12285

Added support for suffix support in variable names, for example:
```
${FOO_1}
${BAR_1:-with-default}
```


ndeloof (approved):

MikaelElkiaer:
@idsulik Looks great.

Just to be clear, this will also work for `${FOO1}`, right?

idsulik:
@MikaelElkiaer thank you! yes it will work for ${FOO1} as well

PR #713, opened 2024-11-22
feat: Add label_file support to service

idsulik:
Issue: https://github.com/docker/compose/issues/12311


idsulik:
@ndeloof hi! What do you think about the PR, should I continue implementing the feature like I do(it uses the same way as we load env_file) or should I do it in a different way?

ndeloof:
Sounds good to me. I'd prefer we can rely on docker/cli parser for this file so we don't have 2 implementations, but this would mean we resolve labels in docker/compose, and on the other hand format is pretty minimalist, so let's go this way

ndeloof (commented):
LGTM

in this PR, you're anticipating the need to set a label_file as `required`. Do you have a use-case in mind ? We introduced this for `env_file` as a workaround for corner-case usages, but I'd prefer we don't make things unnecessary complicated if we don't have to, at least for a first iteration.

idsulik:
> Do you have a use-case in mind

I did it to follow the current logic for the env_file. removed it + added tests

schema/compose-spec.json
```diff
@@ -866,7 +867,34 @@
...
+            ]
+          }
+        }
+      ]
+    },
+
+    "label_file": {
+      "oneOf": [
+        {"type": "string"},
+        {
```
ndeloof, line 888: Sounds to me this is a premature feature to add support for `required` here. `label_file` should be declared at most a `string|array of string` for user convenience

idsulik: @ndeloof agree, fixed

transform/labelfile.go
```diff
@@ -0,0 +1,55 @@
...
+		for i, e := range v {
+			v[i] = transformLabelFileValue(e)
+		}
+		return v, nil
+	default:
+		return nil, fmt.Errorf("%s: invalid type %T for label_file", p, v)
+	}
+}
+
+func transformLabelFileValue(data any) any {
```
ndeloof, line 41: not relevant anymore

loader/full-struct_test.go
```diff
@@ -217,10 +217,24 @@ func services(workingDir, homeDir string) types.Services {
 			Ipc:      "host",
 			Uts:      "host",
 			Labels: map[string]string{
+				"FOO":                     "foo_from_label_file",
```
ndeloof, line 220: FYI this huge test we inherited from legacy compose parser in docker/cli.
As it is painful to maintain, we prefer to have individual test checking specific attribute is well supported (which you also provided 🥳)

idsulik: should I revert these changes or keep them?

idsulik:
@ndeloof pushed changes, now label_file has type `[]string`

types/labelfile.go
```diff
@@ -0,0 +1,36 @@
...
+   limitations under the License.
+*/
+
+package types
+
+import (
+	"encoding/json"
+)
+
+type LabelFile struct {
```
ndeloof, line 23: unused ?

idsulik: yes, thank you, removed

schema/compose-spec.json
```diff
@@ -866,9 +867,6 @@
                   "path": {
                     "type": "string"
                   },
-                  "format": {
-                    "type": "string"
-                  },
```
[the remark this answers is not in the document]
ndeloof: sure, but here you removed `format` attribute from `env_file` defintion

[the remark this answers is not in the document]
idsulik: oh, accidentally deleted it, put it back

ndeloof (approved):

PR #714, opened 2024-11-22
introduce marhsaller options and ability to include secrets content

ndeloof:
This allows to export secret content in yaml output, to be consumed by third-party tools which rely on canonical compose model and need all data resolved. Still, default marshalling output does **not** include content as this would introduce risk to leak sensitive data

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -732,16 +732,18 @@ type CredentialSpecConfig struct {
 
 // FileObjectConfig is a config type for a file used by a service
 type FileObjectConfig struct {
-	Name           string            `yaml:"name,omitempty" json:"name,omitempty"`
-	File           string            `yaml:"file,omitempty" json:"file,omitempty"`
-	Environment    string            `yaml:"environment,omitempty" json:"environment,omitempty"`
-	Content        string            `yaml:"content,omitempty" json:"content,omitempty"`
-	External       External          `yaml:"external,omitempty" json:"external,omitempty"`
-	Labels         Labels            `yaml:"labels,omitempty" json:"labels,omitempty"`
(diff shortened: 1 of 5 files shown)


glours (approved):
LGTM

types/types.go
```diff
@@ -732,16 +732,17 @@ type CredentialSpecConfig struct {
...
-	Labels         Labels            `yaml:"labels,omitempty" json:"labels,omitempty"`
-	Driver         string            `yaml:"driver,omitempty" json:"driver,omitempty"`
-	DriverOpts     map[string]string `yaml:"driver_opts,omitempty" json:"driver_opts,omitempty"`
-	TemplateDriver string            `yaml:"template_driver,omitempty" json:"template_driver,omitempty"`
-	Extensions     Extensions        `yaml:"#extensions,inline,omitempty" json:"-"`
+	Name            string            `yaml:"name,omitempty" json:"name,omitempty"`
+	File            string            `yaml:"file,omitempty" json:"file,omitempty"`
+	Environment     string            `yaml:"environment,omitempty" json:"environment,omitempty"`
+	Content         string            `yaml:"content,omitempty" json:"content,omitempty"`
+	marshallContent bool              // configure marshalling to include Content - excluded by default to prevent sensitive data leaks
```
glours, line 739: Should be public, no?
```suggestion
	MarshallContent bool              // configure marshalling to include Content - excluded by default to prevent sensitive data leaks
```

ndeloof: nope, so that the only way to set this is to use the mashaller option (this is just an implementation detail here, as yaml.Marshall does not allow passing options or custom marshallers)

PR #715, opened 2024-11-27
introduce sync+exec watch action

ndeloof:
introduce ability to configure a sync+exec action, to execute arbitrary command after watch sync updated container

see https://github.com/docker/compose/issues/12318


glours (approved):
LGTM

PR #716, opened 2024-11-29
Fix uniqueItems property defined on object

AnaMManolache:
The `uniqueItems` property is defined for an object, `extra_hosts`, instead of an array.

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -935,7 +935,9 @@
                 },
                 {
                   "type": "array",
-                  "items": {},
+                  "items": {
+                    "type": "string"
+                  },
                   "uniqueItems": false
                 }
               ]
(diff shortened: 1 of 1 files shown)


schema/compose-spec.json
```diff
@@ -918,9 +918,17 @@
...
-              "type": ["string", "array"]
-            },
-            "uniqueItems": false
+              "oneOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "array",
+                  "items": {},
```
ndeloof, line 927: when set with an array, items are actually of type string. Schema should reflect this constraint

ndeloof (approved):

schema/compose-spec.json
AnaMManolache: I've added this constraint now

PR #718, opened 2024-12-09
extending a service, dependencies must be excluded

ndeloof:
exclude `depends_on` and `volumes_from` when extending a service, as documented by https://docs.docker.com/compose/how-tos/multiple-compose-files/extends/#exceptions-and-limitations

closes https://github.com/docker/compose/issues/12360

### DIFF
--- a/loader/extends.go
+++ b/loader/extends.go
@@ -27,6 +27,10 @@ import (
 	"github.com/compose-spec/compose-go/v2/types"
 )
 
+// as we use another service definition by `extends`, we must exclude attributes which creates dependency to another service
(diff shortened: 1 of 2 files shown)


loader/extends.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/compose-spec/compose-go/v2/types"
 )
 
+var exclusions = []string{"extends", "depends_on", "volumes_from"}
```
idsulik, line 30: Maybe it'll be better to add a comment with an explanation of why we exclude these properties, link to the doc https://docs.docker.com/compose/how-tos/multiple-compose-files/extends/#exceptions-and-limitations, or short explanation. I think it'll help in the future to understand what's going on)

PR #727, opened 2024-12-30
Update readme example to use non-deprecated ProjectOptions.LoadProject

moistpope:
[From the docs:](https://pkg.go.dev/github.com/compose-spec/compose-go/v2@v2.4.6/cli#ProjectOptions.LoadProject)

> ProjectFromOptions load a compose project based on command line options 
> Deprecated: use ProjectOptions.LoadProject or ProjectOptions.LoadModel

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ func main() {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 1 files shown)


glours (approved):
LGTM, thanks

PR #729, opened 2025-01-22
allow additional_context to target another service

ndeloof:
Introduce support for `service:xx` syntax defining an additional context.

this will be used to explicitly request another service image being used as source for another service. This is a long term issue inherited from compose v1 sequential build logic: some users declare image built by another service as `FROM` instruction in another service's Dockerfile. They will be able to keep this working by adding an explicit additional_context as `service: baseapp` and let buildkit manage best possible execution of the build.

see https://github.com/docker/buildx/blob/v0.8.0-rc1/docs/reference/buildx_bake.md#using-a-result-of-one-target-as-a-base-image-in-another-target

illustration with docker/compose : https://github.com/docker/compose/pull/12485

### DIFF
--- a/loader/validate.go
+++ b/loader/validate.go
@@ -28,7 +28,7 @@ import (
 
 // checkConsistency validate a compose model is consistent
 func checkConsistency(project *types.Project) error {
-	for _, s := range project.Services {
+	for name, s := range project.Services {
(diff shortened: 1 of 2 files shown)


ndeloof:
cc @crazy-max

glours (approved):
Sounds good to me

PR #731, opened 2025-01-22
introduce gw_priority

ndeloof:
implement https://github.com/compose-spec/compose-spec/pull/552

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -320,7 +320,8 @@
                             "^.+$": {"type": ["string", "number"]}
                           }
                         },
-                        "priority": {"type": "number"}
+                        "priority": {"type": "number"},
+                        "gw_priority": {"type": "number"}
                       },
                       "additionalProperties": false,
                       "patternProperties": {"^x-": {}}
--- a/types/types.go
+++ b/types/types.go
@@ -438,13 +438,14 @@ type PlacementPreferences struct {
 
 // ServiceNetworkConfig is the network configuration for a service
 type ServiceNetworkConfig struct {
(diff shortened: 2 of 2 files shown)


ndeloof:
cc @robmry

robmry (approved):
Oh, thank you! I was about to have a look at this one ... but only got as far as realising I should've updated compose-spec's `schema/compose-spec.json` in https://github.com/compose-spec/compose-spec/pull/552, I'll raise a new PR for that. (Also for `EnableIPv4` in https://github.com/compose-spec/compose-spec/pull/551, but that's not merged yet.)

LGTM - although perhaps it's worth adding to `full-example.yml` / `full-struct_test.go`?

ndeloof:
@robmry we use a bot to sync schema changes made in compose-go into the spec
About full-struct_test.go, we avoid using this huge beast for new feature, and prefer dedicated tests per feature - but for such a trivial thing as a plain yaml mapping we don't require one

robmry:
Ah, got it! Thank you.

ndeloof:
adding support for this in docker/compose is a bit challenging as we can't get API 1.48 from any tag, and depending on `master` bring dependency updates that can create troubles.

robmry:
> adding support for this in docker/compose is a bit challenging as we can't get API 1.48 from any tag, and depending on `master` bring dependency updates that can create troubles.

I expect we'll create a 28.x branch from master, with a release candidate label, in the next week or two ... but it'd be good to find the problems sooner rather than later. Would it be possible to give it an early run in a PR that's not mergeable until the tag's available?

robmry:
Oh, I see - thank you.

PR #736, opened 2025-02-12
refactor: migrate golang.org/x/exp to stds

### LINKED ISSUE #735: Remove usage of x/exp
Imports from `golang.org/x/exp` seem to be some old relics and should be replaced with stdlib.

morlay:
fix #735


types/project.go
```diff
@@ -120,21 +121,21 @@ func (p *Project) ServicesWithBuild() []string {
 	servicesBuild := p.Services.Filter(func(s ServiceConfig) bool {
 		return s.Build != nil && s.Build.Context != ""
 	})
-	return maps.Keys(servicesBuild)
+	return slices.AppendSeq(make([]string, 0), maps.Keys(servicesBuild))
```
morlay, line 124: `slices.Collect` will return nil which break testing.

ndeloof: Why is `AppendSeq` required here, as maps.Keys already return a `[]string`?

morlay: it returns iter.Seq[string]

ndeloof:
compose-go declares requirement on go 1.21, while those func have been introduced in 1.23

types/project.go
ndeloof: testing uses `assert.DeepEqual(t, []string{}, p.ServicesWithBuild())`, you can just switch to `assert.Equal(t, 0, len(p.ServicesWithBuild())` as nil and empty slice should be considered equivalent

ndeloof: done in https://github.com/compose-spec/compose-go/pull/744

ndeloof: not needed anymore AFAICT
otherwise, fix in test should be easy as demonstrated on https://github.com/compose-spec/compose-go/pull/744/files#diff-5fe049f199cfa53efadb44e5077186d86422bd7b585f333dae441b6c60f1eef2L312-R311

PR #738, opened 2025-02-13
Manage secrets|configs.mode as an octal string||number

ndeloof:
A dedicated decoder is required for services.secrets.mode as octal value can be either yaml octal or a plain string

Closes https://github.com/moby/moby/issues/49404


thaJeztah (approved):
not _deeply_ familiar with the codebase, but looks solid, thanks!


LGTM

glours (approved):

PR #740, opened 2025-02-17
feat: Add pattern matching and multiple paths to watch configuration

jhrotko:
This PR enhances the watch configuration by introducing support for multiple paths and pattern matching under `services.develop.watch.path`. This update allows more flexibility in defining file-watching behavior for `rebuild` and `restart` actions.

Example:

```
   1   │ services:
   2   │   api:
   3   │     image: avatars-api
   4   │     build:
   5   │       context: .
   6   │       dockerfile: ./deploy/api.dockerfile
   7   │     ports:
   8   │       - 5734:80
   9   │     develop:
  10   │       watch:
  11   │         - path: api/*.txt
  12   │           action: rebuild
  13   │        - path: 
  14   │           - api
  15   │           - other
  16   │          action: restart
```

This PoC is based on the request of [#12544](https://github.com/docker/compose/issues/12544).
As discussed in this issue, given the restrictions of `sync` actions that require a 1:1 relation between `path` and `target`, I would propose to move this functionality to a new keyword `paths` that would be available for `rebuild` and `restart` actions only. 

Related compose [PR ](https://github.com/docker/compose/pull/12557)

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -3096,7 +3096,7 @@ services:
 		assert.DeepEqual(t, *frontend.Develop, types.DevelopConfig{
 			Watch: []types.Trigger{
 				{
-					Path:   []string{"./webapp/html"},
+					Path:   "./webapp/html",
 					Action: types.WatchActionSync,
 					Target: "/var/www",
 					Ignore: []string{"node_modules/"},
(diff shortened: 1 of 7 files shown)


ndeloof:
While I can see some benefits having support for multiple values in `path` I can't see how this would help with https://github.com/docker/compose/issues/12544 and pattern support, can you please elaborate ?

jhrotko:
> While I can see some benefits having support for multiple values in `path` I can't see how this would help with [docker/compose#12544](https://github.com/docker/compose/issues/12544) and pattern support, can you please elaborate ?

if you take a look at the original issue, the user is requesting for both changes. On a first approach I thought they were hand in hand however with the latest changes is a nice to have. Looking at some existing watch configurations a lot of users are repeating actions (ex.: rebuild) with a different path and I think it would be a nice addition to add multiple path to each action in order to reduce repetition in a compose file.

PR #743, opened 2025-02-20
remove obsolete functions

ndeloof:
see https://github.com/compose-spec/compose-go/issues/742

### DIFF
--- a/loader/extends_test.go
+++ b/loader/extends_test.go
@@ -274,7 +274,7 @@ services:
 	assert.NilError(t, os.WriteFile(filepath.Join(tmpdir, "compose.yaml"), []byte(rootYAML), 0o600))
 
 	extendsCount := 0
-	actual, err := Load(types.ConfigDetails{
+	actual, err := LoadWithContext(context.Background(), types.ConfigDetails{
 		WorkingDir: tmpdir,
 		ConfigFiles: []types.ConfigFile{{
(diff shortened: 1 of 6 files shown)


glours (approved):
We'll need to do a major version as some downstream project may still use the `Load` function, no? WDYT @ndeloof ?

ndeloof:
Load has been deprecated since 1.18 (https://github.com/compose-spec/compose-go/commit/ad9be94130b3c6b06b4827ad5b062f966dcb8dae), so they have been warned!

alexellis:
Thanks for looking at this.

Isn't there a way to mark a method as deprecated in Go through godoc or similar?

https://github.com/compose-spec/compose-go/pull/743/files#diff-8e9af1e483632eac17610237caaa5f879ea7bfb3e1e1ea228e45a17c770b9432L260

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

I didn't see that on the ParseYAML function.

ndeloof:
This one was not marker as deprecated as it's just a legacy (not used by the canonical loading sequence as implemented in docker/compose) we forgot to tag as deprecated/remove

PR #744, opened 2025-02-20
bump golangci-lint

ndeloof:
Use latest golangci-lint and align linters with those set for docker/compose

Require go 1.23 as go 1.24 has been released (https://go.dev/blog/go1.24) and supported version are now 1.23 and 1.24, so doesn't make sense we rely on 1.21 which blocks https://github.com/compose-spec/compose-go/pull/736


glours (approved):
LGTM

PR #745, opened 2025-02-21
introduce pull policy "refresh"

ndeloof:
introduce a new pull policy to refresh image based on last pulled time

to be debated: I selected `refresh` but this might not be the best verb for this feature, I'm open to your proposals :)

see https://github.com/docker/compose/issues/12563


types/types.go
```diff
@@ -215,6 +217,8 @@ const (
 	PullPolicyMissing = "missing"
 	// PullPolicyBuild force building images
 	PullPolicyBuild = "build"
+	// PullPolicyRefresh checks image needs to be updatedq
```
laurazard, line 220:
```suggestion
	// PullPolicyRefresh checks if image needs to be updated
```

glours (approved):
Sounds good to me

PR #746, opened 2025-02-23
fix loading gw_priority

ndeloof:
Forgot to run `make deepcopy` - I always do adding new attributes. So I added a validation step to check this in CI, also updated make targets to run commands in a container vs local tools installation

closes https://github.com/docker/compose/issues/12574


glours (approved):
LGTM

PR #753, opened 2025-03-13
Report file being process in dotenv parsing error

ndeloof:
Include file name being processed as godotenv parser returns error

closes https://github.com/docker/compose/issues/12614


hanikesn:
Thank you!

PR #758, opened 2025-03-27
introduce external services

ndeloof:
https://docker.atlassian.net/browse/APCLI-1065


glours (approved):
👍 simple and effective enough first iteration

rcjsuen:
In light of https://github.com/compose-spec/compose-spec/issues/138 and https://github.com/compose-spec/compose-go/pull/776, could we also add something in the `description` for this new attribute while it is fresh in our minds?

glours:
@rcjsuen I'll do it

PR #760, opened 2025-03-28
Add support for wildcards selecting compose files

ndeloof:
fixes https://github.com/docker/compose/issues/10857


glours:
I’m wondering whether we should support this in compose-go, or if it should be handled on the client side instead?
If we decide to handle it within compose-go, I think we need to have a clear rule in the specification to ensure predictable behavior—something like loading files in alphabetical order.

ndeloof:
> I’m wondering whether we should support this in compose-go, or if it should be handled on the client side instead?

I looked into adding this to docker/compose first, but this is way simpler here.

> I think we need to have a clear rule in the specification to ensure predictable behavior—something like loading files in alphabetical order.

Sure, here I'm using `slices.Sort` and this has to be well documented in the spec

PR #766, opened 2025-04-05
fix: Prevent panic when merging IPAM config

idsulik:
Prevents a panic by adding type checks to ensure both base and override IPAM configs are lists. Returns a clear error if the input types are invalid.

Fixes: https://github.com/docker/compose/issues/12709


idsulik:
@ndeloof hi! I see the issue was closed, but I think it'll be better to fix the panic itself as well

ndeloof:
@idsulik https://github.com/compose-spec/compose-go/pull/767 was merged with a comparable fix, sorry I didn't noticed you already had one ready for this bug

PR #767, opened 2025-04-08
prevent panic merging ipam configs

ndeloof:
fixes https://github.com/docker/compose/issues/12709

Looking into this, I noticed we try to merge ipam config arrays, but doing so we assume we have to merge by same indexes - this is undocumented and probably unexpected. To be revisited


ndeloof:
indeed, didn't noticed this existing PR.
Let me use this one to investigate a fix for incorrect behavior of merge based on index in list vs content

glours (approved):

PR #770, opened 2025-04-14
types: add volume type of api-socket

stevvooe:
Adds definitions for https://github.com/compose-spec/compose-spec/pull/587.


ndeloof:
closing as we introduced `use_api_socket`

PR #786, opened 2025-05-12
provider.options can be an array

ndeloof:
This allows to pass:
```yaml
services:
  cat:
    provider:
      type: schrodinger
      options:
        health:
           - dead
           - alive
```


glours (approved):
👍

PR #798, opened 2025-06-18
Remove support for nvidia driver capabilities

elezar:
This chanage updates the device spec to allow for a driver OR capabilities to be specified. It also removes the capabilities from the gpu spec and indicates that the gpu capability is implied.

### DIFF
--- a/schema/compose-spec.json
+++ b/schema/compose-spec.json
@@ -1226,8 +1226,17 @@
         },
         "additionalProperties": false,
         "patternProperties": {"^x-": {}},
-        "required": [
-          "capabilities"
+        "oneOf": [
+          {
+            "required": [
+              "capabilities"
+            ]
+          },
+          {
+            "required": [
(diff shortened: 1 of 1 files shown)


ndeloof:
>  It also removes the capabilities from the gpu spec and indicates that the gpu capability is implied.

true, still other capabilities could be required to select the adequate device.
Also, being version-less, compose-spec doesn't allow attributes to be removed from the spec.

elezar:
> > It also removes the capabilities from the gpu spec and indicates that the gpu capability is implied.
> 
> true, still other capabilities could be required to select the adequate device. Also, being version-less, compose-spec doesn't allow attributes to be removed from the spec.

Thanks for pointing this out. I have updated the PR to not remove the field, but to update the documentation for the field to not refer to the nvidia-specific capabilities.

ndeloof (approved):

PR #808, opened 2025-07-14
feat: add extensible transform API for custom field transformations

OrlovEvgeny:
This enables users to extend compose-go with custom protocols and validation                                                                                logic without forking the project, addressing the need for custom port  protocols like http/https and domain name support

- Add RegisterDefaultValue() functions  
- Add EnhancedPortDefaults as example of extensible port transformation 
- Maintain backward compatibility with existing code

example: 
```go
// EnhancedPortDefaults is an example of an enhanced port defaults function
// that supports additional protocols and domain names in the published field.
// This function can be used as a replacement for the default portDefaults
// by registering it with RegisterDefaultValue.
//
// Example usage:
//   transform.RegisterDefaultValue("services.*.ports.*", transform.EnhancedPortDefaults)
//
// This function supports:
// - Additional protocols: "http", "https", "tcp", "udp"
// - Domain names in the published field (e.g., "example.com:80")
// - All existing functionality of the original portDefaults
func EnhancedPortDefaults(data any, _ tree.Path, _ bool) (any, error) {
	switch v := data.(type) {
	case map[string]any:
		// Set default protocol if not specified
		if _, ok := v["protocol"]; !ok {
			v["protocol"] = "tcp"
		}
		// Set default mode if not specified
		if _, ok := v["mode"]; !ok {
			v["mode"] = "ingress"
		}
		// Additional logic can be added here to handle domain names
		// in the published field or other custom transformations
		return v, nil
	default:
		return data, nil
	}
}
```

Related issue: https://github.com/psviderski/uncloud/issues/44

### DIFF
--- a/transform/canonical.go
+++ b/transform/canonical.go
@@ -22,9 +22,10 @@ import (
 	"github.com/compose-spec/compose-go/v2/tree"
 )
 
-type transformFunc func(data any, p tree.Path, ignoreParseError bool) (any, error)
+// Func is a function that can transform data at a specific path
+type Func func(data any, p tree.Path, ignoreParseError bool) (any, error)
 
-var transformers = map[tree.Path]transformFunc{}
+var transformers = map[tree.Path]Func{}
 
 func init() {
 	transformers["services.*"] = transformService
--- a/transform/defaults.go
+++ b/transform/defaults.go
@@ -20,14 +20,20 @@ import (
 	"github.com/compose-spec/compose-go/v2/tree"
 )
 
-var defaultValues = map[tree.Path]transformFunc{}
+// DefaultValues contains the default value transformers for compose fields
+var DefaultValues = map[tree.Path]Func{}
 
 func init() {
-	defaultValues["services.*.build"] = defaultBuildContext
-	defaultValues["services.*.secrets.*"] = defaultSecretMount
-	defaultValues["services.*.ports.*"] = portDefaults
-	defaultValues["services.*.deploy.resources.reservations.devices.*"] = deviceRequestDefaults
-	defaultValues["services.*.gpus.*"] = deviceRequestDefaults
+	DefaultValues["services.*.build"] = defaultBuildContext
+	DefaultValues["services.*.secrets.*"] = defaultSecretMount
+	DefaultValues["services.*.ports.*"] = portDefaults
+	DefaultValues["services.*.deploy.resources.reservations.devices.*"] = deviceRequestDefaults
+	DefaultValues["services.*.gpus.*"] = deviceRequestDefaults
+}
+
+// RegisterDefaultValue registers a custom transformer for the given path pattern
+func RegisterDefaultValue(path string, transformer Func) {
+	DefaultValues[tree.Path(path)] = transformer
 }
 
 // SetDefaultValues transforms a compose model to set default values to missing attributes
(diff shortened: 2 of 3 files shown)


transform/defaults.go
```diff
@@ -20,7 +20,11 @@ import (
 	"github.com/compose-spec/compose-go/v2/tree"
 )
 
-var defaultValues = map[tree.Path]transformFunc{}
+// DefaultValues contains the default value transformers for compose fields
+var DefaultValues = map[tree.Path]TransformFunc{}
+
+// For backward compatibility
+var defaultValues = DefaultValues
```
ndeloof, line 27: not needed, as a non-public field there's no backward compatibility to be addressed here

transform/defaults.go
```diff
@@ -30,6 +34,20 @@ func init() {
 	defaultValues["services.*.gpus.*"] = deviceRequestDefaults
 }
 
+// RegisterDefaultTransformer registers a custom transformer for the given path pattern
+func RegisterDefaultTransformer(path string, transformer TransformFunc) {
```
ndeloof, line 38: `RegisterDefaultTransformer` sounds weird to me. What about `RegisterDefaultValue`

transform/defaults.go
```diff
@@ -30,6 +34,20 @@ func init() {
...
+	DefaultValues[tree.Path(path)] = transformer
+}
+
+// GetDefaultTransformers returns a copy of the current default transformers
+func GetDefaultTransformers() map[string]TransformFunc {
```
ndeloof, line 43: why have this function while `DefaultValues` is public ?

ndeloof (approved):

OrlovEvgeny:
Hey @ndeloof 
> Can you please squash your commits, rebase and sign-off, so we can merge ?
Done

OrlovEvgeny:
I see,It seems something broke after I rebased main
let me fix it

OrlovEvgeny:
@ndeloof fixed

transform/canonical.go
```diff
@@ -20,7 +20,11 @@ import (
 	"github.com/compose-spec/compose-go/v2/tree"
 )
 
-type transformFunc func(data any, p tree.Path, ignoreParseError bool) (any, error)
+// TransformFunc is a function that can transform data at a specific path
+type TransformFunc func(data any, p tree.Path, ignoreParseError bool) (any, error)
+
+// For backward compatibility
```
ndeloof, line 26: unnecessary

OrlovEvgeny:
Linter complains about "stuttering" when using type as `transform.TransformFunc`
https://github.com/compose-spec/compose-go/actions/runs/16293608125/job/46012331031#step:4:33

I refactored some naming:
renamed `TransformFunc` to `Func` in transform/canonical.go
what do you think about it?

glours (commented):
Sounds good!
I wonder if we should not add a test, to avoid breaking change on the feature and give an exemple for developers who want to use it, WDYT?

OrlovEvgeny:
> Sounds good! I wonder if we should not add a test, to avoid breaking change on the feature and give an exemple for developers who want to use it, WDYT?

Yes, I think that's a good idea 

I added a test and an example function `exampleEnhancedPortDefaults`

glours (approved):
LGTM, thanks a lot for this contribution 🙏

PR #828, opened 2025-10-28
Fix models support with compose include

ndeloof:
based on https://github.com/compose-spec/compose-go/pull/824


ndeloof:
@doringeman need to bump compose-go dependency in docker/compose

PR #834, opened 2025-11-04
introduce no-cache-filter

ndeloof:
aligned with https://docs.docker.com/build/cache/invalidation/#run-instructions

see https://github.com/docker/compose/issues/13335


glours (approved):
LGTM, could be merged when the conflict will be resolved

PR #860, opened 2026-04-04
fix: ensure deterministic variable interpolation

yunus25jmi1:
## Description

This fixes the non-deterministic behavior when multiple required variables (e.g., ${TIMEZONE:?}, ${TRAEFIK_ACME_PATH:?}) are missing in docker-compose.yaml.

## Root Cause

Go's map iteration order is randomized. When iterating over service environment variables or volumes during interpolation, different keys can be processed in different orders on each run. This causes the error message for missing required variables to vary between runs.

## Fix

Added sort.Strings(keys) before map iteration in two places:
1. In Interpolate() function - sorts top-level config keys
2. In recursiveInterpolate() function - sorts nested map keys

This ensures deterministic variable interpolation order, so the same missing variable is always reported first.

## Testing

- go build: passes
- go test ./interpolation/: all 7 tests pass
- go vet: passes

## Related

- Fixes https://github.com/docker/compose/issues/13712

### DIFF
--- a/interpolation/interpolation.go
+++ b/interpolation/interpolation.go
@@ -17,6 +17,7 @@
 package interpolation
 
 import (
+	"sort"
 	"errors"
 	"fmt"
 	"os"
@@ -58,7 +59,13 @@ func Interpolate(config map[string]interface{}, opts Options) (map[string]interf
 
 	out := map[string]interface{}{}
 
-	for key, value := range config {
+	keys := make([]string, 0, len(config))
+	for key := range config {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		value := config[key]
 		interpolatedValue, err := recursiveInterpolate(value, tree.NewPath(key), opts)
 		if err != nil {
 			return out, err
@@ -88,7 +95,13 @@ func recursiveInterpolate(value interface{}, path tree.Path, opts Options) (inte
 
 	case map[string]interface{}:
 		out := map[string]interface{}{}
-		for key, elem := range value {
+		keys := make([]string, 0, len(value))
+		for key := range value {
+			keys = append(keys, key)
+		}
+		sort.Strings(keys)
+		for _, key := range keys {
+			elem := value[key]
 			interpolatedElem, err := recursiveInterpolate(elem, path.Next(key), opts)
 			if err != nil {
 				return nil, err
--- a/interpolation/interpolation.go
+++ b/interpolation/interpolation.go
@@ -17,10 +17,11 @@
 package interpolation
 
 import (
-	"sort"
 	"errors"
 	"fmt"
+	"maps"
 	"os"
+	"slices"
 
 	"github.com/compose-spec/compose-go/v2/template"
 	"github.com/compose-spec/compose-go/v2/tree"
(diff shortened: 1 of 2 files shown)


interpolation/interpolation.go
```diff
@@ -58,7 +59,13 @@ func Interpolate(config map[string]interface{}, opts Options) (map[string]interf
 
 	out := map[string]interface{}{}
 
-	for key, value := range config {
+	keys := make([]string, 0, len(config))
```
thaJeztah: perhaps the slices and maps packages could be used here; something like; 

```go
keys := maps.Keys(m)
slices.Sort(keys)
```

thaJeztah: Oh! That won't work because maps.Keys returns an iterator; this probably works though;

```go
keys := slices.Sorted(maps.Keys(config))
```

yunus25jmi1:
Thanks for the suggestion @thaJeztah! I've updated the code to use `slices.Sorted(maps.Keys(config))` as you recommended - this is much more idiomatic Go.

Changes:
- Replaced manual slice creation + sort.Strings with slices.Sorted(maps.Keys())
- Added "slices" and "maps" imports
- Applied to both locations (Interpolate and recursiveInterpolate functions)

All tests pass.

thaJeztah:
Ah, nice, thanks!

There was nothing wrong with your original implementation (and in some cases open-coding these things could still be used for performance critical paths) but the slices and maps packages can make the code cleaner, easier to understand and maintain, which is a big plus :-)

yunus25jmi1:
@ndeloof review the PR.

thaJeztah:
I think Nicolas is off for a few days, but a maintainer will probably review this change after the weekend / after Easter.

FWIW; (I haven't dug deep into the code) I think this change would make failures more deterministic, but I _think_ there may still be situations where it could fail even though technically it's valid; this PR sorts the env-vars in a deterministic way, but expansion may fail if an expansion depends on another env-var that's sorted "later"?

Haven't tested it yet, but considering something like this;

```yaml
- ${AAAA:-${BBBB:-${ZZZZ}}}
- ${BBBB:-${ZZZZ}}
- ZZZZ=default
```

Those are sorted, but if those are interpolated in alphabetic order, they would likely still fail.

Basically; 

- `AAAA` depends on `BBBB`
- `BBBB` depends on `ZZZZ`
- `ZZZZ` has a default value

So the _correct_ order would be to expand them in topological order;

- `ZZZZ` first
- `BBBB` second
- `AAAA` last

yunus25jmi1:
Thanks @thaJeztah for the suggestion! I've updated the code to use `slices.Sorted(maps.Keys(config))` as you recommended.

I also investigated your comment about nested variable dependencies (e.g., `${AAAA:-}`). The template package already handles this correctly through recursive substitution - when a default value contains another variable, it recursively resolves the inner variables first before using them as defaults for outer ones. This provides the correct topological ordering automatically.

I've added test cases to verify this behavior. Let me know if there are any other changes or suggestions needed before the PR can be merged!

yunus25jmi1:
Hi @ndeloof @thaJeztah and maintainers,

This PR is ready for review. All feedback from @thaJeztah has been addressed:
- Updated to use idiomatic Go (slices.Sorted(maps.Keys()))
- Added test cases for nested variable expansion

This fix is aligned with docker/compose PR #13713 (https://github.com/docker/compose/pull/13713) which addresses the same non-deterministic variable interpolation issue. Once this PR is merged, we can update docker/compose to use the new compose-go version with the fix.

Please let me know if any additional changes are needed!

Signed-off-by: Md Yunus <admin@yunuscollege.eu.org>

PR #866, opened 2026-04-17
Jobs

ndeloof:
implements https://github.com/compose-spec/compose-spec/pull/642

The container configuration is extracted as a layered specification — `ContainerSpec` (anything that runs a container: services, jobs, `pre_start` init containers) / `WorkloadSpec` (services and jobs: build, depends_on, expose, healthcheck, ports, stdin_open, tty) / service-only attributes. `pre_start` hooks carry the full container specification (compose-spec#656), and the canonicalization/path-resolution/interpolation registries follow the same layering (which also fixes interpolation casts never applying to jobs).

---

**Suggested review order** — the history reads linearly, no commit revisits an earlier one's design:

1. `test(loader): golden-file harness` — pins today's loader output before anything moves;
2. `refactor(types): extract ContainerSpec and WorkloadSpec` — the layering, no behavior change;
3. `feat: pre_start hooks carry the full container specification` — hooks become a ContainerSpec (compose-spec#656);
4. `feat: introduce the jobs top-level element` — jobs with triggers (schedule + tri-state manual), full pipeline support;
5. `feat: introduce WithSelectedJob` — project narrowing for manual job execution;
6. `feat: resolve pre_start inheritance at load time` — the model carries resolved hooks, idempotently across config round-trips (relies on the strictly-identical-entry merge dedup of #924, now on main).


types/project.go
```diff
@@ -569,6 +608,40 @@ func (p *Project) WithoutUnresolvedOptionalDependencies() *Project {
...
+			enabled, err := p.WithProfiles(append(append([]string{}, p.Profiles...), disabled.Profiles...))
+			if err != nil {
+				return nil, err
+			}
+			return enabled.WithSelectedJob(name, options...)
+		}
+		return nil, fmt.Errorf("no such job: %s", name)
+	}
+
+	var deps []string
```
glours: In `WithSelectedJob` (types/project.go:628), we build `deps` straight from `job.DependsOn` and pass it to `WithSelectedServices`, but that only resolves against `p.Services`. Since jobs can depend on other jobs (per the spec — `workload_spec.md`: "Declared on a job, `depends_on` can reference both services and other jobs"), if `job.DependsOn` contains a job name we get `no such service: <name>` instead of pulling in the services that job itself needs.

Repro: job `migrate` depends on job `prep`, `prep` depends on service `db` → `WithSelectedJob("migrate")` errors with `no such service: prep` instead of resolving down to `db`.

I think we need to walk job-typed deps recursively before calling `WithSelectedServices`, something like:

```go
deps := p.resolveJobServiceDeps(job, map[string]bool{name: true})
```

```go
// resolveJobServiceDeps walks a job's DependsOn, following job-typed
// dependencies transitively, and returns the names of the actual services
// required to run it.
func (p *Project) resolveJobServiceDeps(job JobConfig, seen map[string]bool) []string {
	var services []string
	for dep := range job.DependsOn {
		if seen[dep] {
			continue
		}
		seen[dep] = true
		if depJob, ok := p.AllJobs()[dep]; ok {
			services = append(services, p.resolveJobServiceDeps(depJob, seen)...)
			continue
		}
		services = append(services, dep)
	}
	return services
}
```

None of the `WithSelectedJob` tests cover job→job deps yet, might be worth adding one for that case too.

glours (changes requested):
One blocker on WithSelectedJob, see inline comment.

types/project.go
ndeloof: Good catch, fixed. `WithSelectedJob` now resolves `depends_on` transitively through job-typed dependencies: a new `resolveJobServiceDeps` helper walks the dependency graph, following any dependency that names a job (job lookup takes precedence over a same-named service) down to the actual services, with a `seen` set seeded with the selected job to guard against job↔job cycles. The resulting service list is sorted before being handed to `WithSelectedServices` so selection stays deterministic.

Added unit tests in `TestWithSelectedJob`: job→job→service resolution (your repro shape — selecting the top job pulls in `db`) and a job↔job cycle that terminates and still keeps the transitive service dependency; the existing direct job→service case is unchanged.

Squashed into the `feat: introduce WithSelectedJob` commit (now d0e5bba); new head is aecd5f8.

types/project.go
```diff
@@ -569,6 +608,62 @@ func (p *Project) WithoutUnresolvedOptionalDependencies() *Project {
...
+// and a job reference is resolved first. seen guards against dependency
+// cycles between jobs and must contain the starting job's name.
+func (p *Project) resolveJobServiceDeps(job JobConfig, seen map[string]bool) []string {
+	var services []string
+	for dep := range job.DependsOn {
+		if seen[dep] {
+			continue
+		}
+		seen[dep] = true
+		if depJob, ok := p.Jobs[dep]; ok {
```
glours: We should also look to disable jobs, no?
```suggestion
		if depJob, ok := p.AllJobs()[dep]; ok {
```

ndeloof: Right — AllJobs() it is (hoisted once before the recursion rather than rebuilt per call): a profile-disabled dependency is still a job, and treating its name as a service would fail the selection. Added a test where a job depends on a disabled job whose own service dependency must be pulled in (release -> backup[profiles: ops] -> redis). Folded into the WithSelectedJob commit: fcc3454, new head a01842c.

glours (approved):
LGTM

PR #870, opened 2026-05-11
types: add Options field to IPAMConfig

c-tonneslan:
The Compose spec has supported `ipam.options` (driver-specific IPAM options) for a while, and the JSON schema already defines it:

```json
"options": {
  "type": "object",
  "description": "Driver-specific options for the IPAM driver.",
  "patternProperties": {"^.+$": {"type": "string"}}
}
```

But `IPAMConfig` in Go was missing the `Options` field, so it silently dropped any options set in the compose file and they never made it to the Docker daemon (`"Options": null` in `docker network inspect`).

Three changes to fix this:
- Add `Options Mapping` to `IPAMConfig` struct in `types/types.go`
- Update the generated `deriveDeepCopy_50` in `derived.gen.go` to copy the new field (same pattern as other `map[string]string` fields)
- Add assertion to the existing `TestTopLevelNetworks` test that options are parsed and survive a round-trip

Related to docker/compose#13785 — this is the compose-go change needed before the docker/compose side can be fixed.

### DIFF
--- a/loader/tests/networks_test.go
+++ b/loader/tests/networks_test.go
@@ -68,7 +68,7 @@ networks:
     ipam:
       driver: overlay
       options:
-        foo: bar
+        ipam-opt-key: ipam-opt-value
       config:
         - subnet: 172.28.0.0/16
           ip_range: 172.28.5.0/24
@@ -99,7 +99,7 @@ networks:
 		assert.Equal(t, other.Driver, "overlay")
 		assert.DeepEqual(t, other.DriverOpts, types.Options{"foo": "bar", "baz": "1"})
 		assert.Equal(t, other.Ipam.Driver, "overlay")
-		assert.DeepEqual(t, other.Ipam.Options, types.Mapping{"foo": "bar"})
+		assert.DeepEqual(t, other.Ipam.Options, types.Mapping{"ipam-opt-key": "ipam-opt-value"})
 		assert.Equal(t, len(other.Ipam.Config), 2)
 		assert.Equal(t, other.Ipam.Config[0].Subnet, "172.28.0.0/16")
 		assert.Equal(t, other.Ipam.Config[0].IPRange, "172.28.5.0/24")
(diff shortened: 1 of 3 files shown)


glours:
Hey @c-tonneslan 
Thanks for the contribution 🙏 
Could you run `make deepcopy` to regenerate `derived.gen.go` instead of editing it manually, and also sign off your commit, please?

loader/tests/networks_test.go
```diff
@@ -97,6 +99,7 @@ networks:
 		assert.Equal(t, other.Driver, "overlay")
 		assert.DeepEqual(t, other.DriverOpts, types.Options{"foo": "bar", "baz": "1"})
 		assert.Equal(t, other.Ipam.Driver, "overlay")
+		assert.DeepEqual(t, other.Ipam.Options, types.Mapping{"foo": "bar"})
```
glours: It would be better to use dedicated keys to avoid confusion if a test fails.
```suggestion
		assert.DeepEqual(t, other.Ipam.Options, types.Mapping{"ipam-opt-key": "ipam-opt-value"})
```

loader/tests/networks_test.go
glours, line 124: Can we add an assertion to ensure we have a correct deep-copy of the added attributes?
```suggestion
	expect(jsonP)
	
	// Exercise the generated deep-copy path (not covered by marshal/load round-trip).
	copied, err := p.WithProfiles(nil)
	assert.NilError(t, err)
	expect(copied)
```

glours (commented):
LGTM, but it would be great if we could polish the PR a bit before merging.

c-tonneslan:
Done. Updated the test to use `ipam-opt-key`/`ipam-opt-value`, added the `WithProfiles(nil)` assertion to cover the deep-copy path, and signed off the commit.

glours:
@c-tonneslan you still need to [signoff your commits](https://github.com/compose-spec/compose-go/pull/870/checks?check_run_id=75425831288)

glours (approved):

PR #875, opened 2026-05-21
loader: add WithSelectedServices and WithoutUnnecessaryResources options

ndeloof:
## Summary

Expose two new loader options so consumers can express service selection
and resource pruning as part of the standard parsing pipeline, without
having to disable environment resolution and re-invoke
`project.WithServicesEnvironmentResolved` manually after loading.

- `loader.WithSelectedServices([]string)` / `cli.WithSelectedServices(...string)`
  - empty slice keeps all services (no-op).
  - otherwise applies `project.WithServicesEnabled(...)` then
    `project.WithSelectedServices(...)`, restricting the project to the
    requested services and their dependencies.
- `loader.WithoutUnnecessaryResources` / `cli.WithoutUnnecessaryResources`
  - drops networks / volumes / secrets / configs / models that are not
    referenced by services remaining after selection.

Both transformations are applied **before**
`WithServicesEnvironmentResolved` (and the implicit
`WithServicesLabelsResolved`) in `ModelToProject`. Concretely, this means
the loader no longer reads `env_file` / `label_file` entries from disk
for services that have been filtered out -- a `required: true` `env_file`
on an unselected service is no longer fatal.

## Motivation

Today, callers that want to select a subset of services have to do:

```go
// load without resolving env
proj, _ := opts.LoadProject(ctx) // with cli.WithoutEnvironmentResolution
// filter services themselves (outside compose-go)
proj, _ = proj.WithSelectedServices(targets)
// resolve env after filtering
proj, _ = proj.WithServicesEnvironmentResolved(true)
```

This is what docker/compose currently does in
[`pkg/compose/loader.go::postProcessProject`](https://github.com/docker/compose/blob/main/pkg/compose/loader.go)
and in `cmd/compose/compose.go::WithServices`. The two-step approach is
needed to avoid reading `env_file` entries (or failing on missing ones)
for services the caller is not interested in. With this PR the same
result can be expressed in a single call:

```go
proj, _ := opts.LoadProject(ctx, cli.WithSelectedServices(targets...))
```

## Tests

- `TestWithSelectedServicesOption` — empty list keeps all services;
  non-empty restricts; dependencies of selected services are kept.
- `TestWithSelectedServicesOption_SkipsMissingEnvFile` — verifies that a
  `required: true` `env_file` pointing to a missing file on an
  unselected service does NOT fail the load.
- `TestWithoutUnnecessaryResourcesOption` — verifies networks / volumes
  referenced only by filtered-out services are pruned when the option
  is set, kept otherwise.

## Test plan

- [x] `go test ./...` passes locally
- [x] new tests cover the three behaviors above

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -383,6 +383,24 @@ func WithoutEnvironmentResolution(o *ProjectOptions) error {
 	return nil
 }
 
+// WithSelectedServices restricts the loaded project to the given services and their
+// dependencies. An empty list means "all services". When set, services not in the
+// list are dropped from the project before environment resolution, so their
+// `env_file` / `label_file` entries are not loaded from disk.
+func WithSelectedServices(services ...string) ProjectOptionsFn {
+	return func(o *ProjectOptions) error {
+		o.loadOptions = append(o.loadOptions, loader.WithSelectedServices(services))
+		return nil
+	}
+}
+
+// WithoutUnnecessaryResources drops networks/volumes/secrets/configs/models that
+// are not referenced by services remaining after selection.
+func WithoutUnnecessaryResources(o *ProjectOptions) error {
+	o.loadOptions = append(o.loadOptions, loader.WithoutUnnecessaryResources)
+	return nil
+}
+
 // DefaultFileNames defines the Compose file names for auto-discovery (in order of preference)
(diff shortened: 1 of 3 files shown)


glours (approved):
LGTM

PR #880, opened 2026-06-01
feat(types): add IPAMConfig.Options field

ndeloof:
## Summary
- Add an `Options` field of type `Options` (a.k.a. `map[string]string`) to `types.IPAMConfig`, with yaml/json tag `options`
- Regenerate `types/derived.gen.go` to deep-copy the new field
- Add a dedicated loader test (`loader/tests/network_ipam_test.go`) covering `ipam.driver`, `ipam.options` and `ipam.config` round-trip
- Extend the existing `TestLoadV2` fixture with `options` under `with_ipam`

## Why
The Compose spec schema (`schema/compose-spec.json`) already allows `networks.<x>.ipam.options` as a driver-specific options map (analogous to `options:` under `logging:` or `driver_opts:` under a network). However, the Go `IPAMConfig` struct didn't expose this field, so entries provided under `ipam.options` were silently dropped on load and never reached consumers (e.g. docker/compose, which currently can't forward them to the daemon).

This makes `IPAMConfig` consistent with the JSON schema, and with sibling structures like `NetworkConfig.DriverOpts` or `LoggingConfig.Options`.

Reported downstream as docker/compose#13785.

## Test plan
- [x] `go test ./loader/...` — all tests pass, including the new `TestNetworkIPAM`
- [x] New test exercises both YAML and JSON round-trip (via the existing `roundTrip` helper)

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -106,6 +106,8 @@ networks:
       driver: default
       config:
         - subnet: 172.28.0.0/16
+      options:
+        foo: bar
 `
 
 var samplePortsConfig = []types.ServicePortConfig{
@@ -207,6 +209,9 @@ var sampleConfig = types.Config{
 					{
 						Subnet: "172.28.0.0/16",
 					},
+				},
+				Options: types.Options{
+					"foo": "bar",
 				},
 			},
 		},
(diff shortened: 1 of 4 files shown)


glours (approved):
LGTM

PR #884, opened 2026-06-08
Pin image mount digests in WithImagesResolved

ssam18:
docker compose config --resolve-image-digests pins service image tags to digests but leaves image mounts untouched, so a volume of type image keeps a floating tag. This extends WithImagesResolved so it also pins the source of image mounts the same way it pins the service image. The per image resolution moved into a small helper that is reused for both cases, with a test covering an image mount alongside a bind mount that stays untouched. Reported in docker/compose#13827.

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -579,28 +579,50 @@ func (p *Project) WithServicesDisabled(names ...string) *Project {
 // It returns a new Project instance with the changes and keep the original Project unchanged
 func (p *Project) WithImagesResolved(resolver func(named reference.Named) (godigest.Digest, error)) (*Project, error) {
 	return p.WithServicesTransform(func(_ string, service ServiceConfig) (ServiceConfig, error) {
-		if service.Image == "" {
-			return service, nil
-		}
-		named, err := reference.ParseDockerRef(service.Image)
-		if err != nil {
-			return service, err
+		if service.Image != "" {
+			resolved, err := resolveImageDigest(service.Image, resolver)
+			if err != nil {
+				return service, err
+			}
+			service.Image = resolved
 		}
 
-		if _, ok := named.(reference.Canonical); !ok {
-			// image is named but not digested reference
-			digest, err := resolver(named)
+		for i, volume := range service.Volumes {
+			if volume.Type != VolumeTypeImage || volume.Source == "" {
+				continue
+			}
+			resolved, err := resolveImageDigest(volume.Source, resolver)
 			if err != nil {
 				return service, err
 			}
-			named, err = reference.WithDigest(named, digest)
-			if err != nil {
-				return service, err
-			}
+			volume.Source = resolved
+			service.Volumes[i] = volume
 		}
-		service.Image = named.String()
+
 		return service, nil
 	})
+}
+
+// resolveImageDigest returns image pinned to a digest. If image is not already a
+// digested (canonical) reference, the digest is obtained from resolver.
+func resolveImageDigest(image string, resolver func(named reference.Named) (godigest.Digest, error)) (string, error) {
+	named, err := reference.ParseDockerRef(image)
+	if err != nil {
+		return image, err
+	}
+
+	if _, ok := named.(reference.Canonical); !ok {
+		// image is named but not digested reference
+		digest, err := resolver(named)
+		if err != nil {
+			return image, err
+		}
+		named, err = reference.WithDigest(named, digest)
+		if err != nil {
+			return image, err
+		}
+	}
+	return named.String(), nil
 }
 
 type marshallOptions struct {
--- a/types/project.go
+++ b/types/project.go
@@ -575,12 +575,29 @@ func (p *Project) WithServicesDisabled(names ...string) *Project {
 	return newProject
 }
 
-// WithImagesResolved updates services images to include digest computed by a resolver function
-// It returns a new Project instance with the changes and keep the original Project unchanged
+// WithImagesResolved updates services to pin both service.Image and image-mount
+// volume sources (type=image) to digests computed by the resolver. It returns a
+// new Project with the changes and keeps the original Project unchanged. Within
+// each service, repeated lookups for the same image reference are deduplicated
+// so callers that hit a registry per resolve aren't charged twice when the
+// service image and an image-mount volume reference the same tag.
 func (p *Project) WithImagesResolved(resolver func(named reference.Named) (godigest.Digest, error)) (*Project, error) {
 	return p.WithServicesTransform(func(_ string, service ServiceConfig) (ServiceConfig, error) {
+		cache := map[string]string{}
+		resolve := func(img string) (string, error) {
+			if r, ok := cache[img]; ok {
+				return r, nil
+			}
+			r, err := resolveImageDigest(img, resolver)
+			if err != nil {
+				return img, err
+			}
+			cache[img] = r
+			return r, nil
+		}
+
 		if service.Image != "" {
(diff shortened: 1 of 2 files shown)


ssam18:
This is the compose-go side of the fix for docker/compose#13827. Once it lands and gets released I will follow up with a compose PR that bumps compose-go so the config and publish commands pick it up. Happy to tweak the helper naming or add more cases if you would prefer.

glours (changes requested):
The `compose-go` change itself looks reasonable to me: `WithImagesResolved` is the right place to extend digest resolution to image mount sources, and the behavior is consistent with how service images are handled today.

That said, I don’t think this is enough to fix the Docker Compose behavior end-to-end. In docker/compose, `--lock-image-digests` and the publish digest override paths currently build an override that only keeps `services.*.image`, so any resolved image mount source would still be dropped there. The `--no-interpolate --resolve-image-digests` path also builds a pseudo-project containing only service images before calling `WithImagesResolved`.

So a compose-go bump alone won’t fully address docker/compose#13827, and if the follow-up Compose PR only bumps compose-go, Dependabot would cover that anyway. We’ll need corresponding docker/compose changes/tests for the lock/publish/no-interpolate paths, or the scope of this PR should be narrowed accordingly.

I’d also like to see a bit more compose-go coverage here:
  - service without `image` but with a `type: image` volume;
  - already canonical image mount source, ensuring the resolver is not called;
  - invalid/unresolvable image mount source.

The `WithImagesResolved` comment should also be updated since it no longer resolves only service images.

types/project_test.go
```diff
@@ -209,6 +209,128 @@ func Test_ResolveImages(t *testing.T) {
...
+	p := &Project{
+		Services: Services{
+			"app": ServiceConfig{
+				Name:  "app",
+				Image: pinned,
+				Volumes: []ServiceVolumeConfig{
+					{
+						Type:   VolumeTypeImage,
+						Source: pinned,
+						Target: "/test_mount",
```
glours, line 295: Same canonical string is used for both `service.Image` and `volume.Source`, so `resolverCalls == 0` can be explained by canonical detection *or* by the per-service cache. A regression that broke canonical detection but kept the cache would still pass.

Suggested fix: drop `service.Image` so only the volume path is exercised

types/project.go
```diff
@@ -575,34 +575,73 @@ func (p *Project) WithServicesDisabled(names ...string) *Project {
...
+		resolve := func(img string) (string, error) {
+			if r, ok := cache[img]; ok {
+				return r, nil
+			}
+			r, err := resolveImageDigest(img, resolver)
+			if err != nil {
+				return img, err
+			}
+			cache[img] = r
+			return r, nil
```
glours: A project with N services referencing the same image triggers N resolver calls. Hoist the cache to WithImagesResolved as a sync.Map

```
func (p *Project) WithImagesResolved(resolver func(named reference.Named) (godigest.Digest, error)) (*Project, error) {                                                                                                                                    
      var cache sync.Map // map[string]string                                                                                                                                                                                                                
      return p.WithServicesTransform(func(_ string, service ServiceConfig) (ServiceConfig, error) {                                                                                                                                                          
          resolve := func(img string) (string, error) {                                                                                                                                                                                                      
              if r, ok := cache.Load(img); ok {                                                                                                                                                                                                              
                  return r.(string), nil                                                                                                                                                                                                                   
              }                                                                                                                                                                                                                                              
              r, err := resolveImageDigest(img, resolver)                                                                                                                                                                                                  
              if err != nil {
                  return img, err                                                                                                                                                                                                                            
              }
              cache.Store(img, r)                                                                                                                                                                                                                            
              return r, nil                                                                                                                                                                                                                                
          }
          // ... rest unchanged
      })
  }
```

types/project.go
```diff
@@ -575,34 +575,73 @@ func (p *Project) WithServicesDisabled(names ...string) *Project {
...
 		}
-		service.Image = named.String()
+
 		return service, nil
 	})
 }
 
+// resolveImageDigest returns image pinned to a digest. If image is not already a
+// digested (canonical) reference, the digest is obtained from resolver.
+func resolveImageDigest(image string, resolver func(named reference.Named) (godigest.Digest, error)) (string, error) {
```
glours: collides visually with the local named declared on line 626. Drop the name in the type literal
```suggestion
func resolveImageDigest(image string, resolver func(reference.Named) (godigest.Digest, error)) (string, error) {
```

types/project.go
```diff
@@ -575,34 +575,73 @@ func (p *Project) WithServicesDisabled(names ...string) *Project {
...
 }
 
-// WithImagesResolved updates services images to include digest computed by a resolver function
-// It returns a new Project instance with the changes and keep the original Project unchanged
+// WithImagesResolved updates services to pin both service.Image and image-mount
+// volume sources (type=image) to digests computed by the resolver. It returns a
+// new Project with the changes and keeps the original Project unchanged. Within
+// each service, repeated lookups for the same image reference are deduplicated
+// so callers that hit a registry per resolve aren't charged twice when the
+// service image and an image-mount volume reference the same tag.
```
glours: Will be true only if we use the cache at the level project not per-service

types/project_test.go
```diff
@@ -209,6 +209,128 @@ func Test_ResolveImages(t *testing.T) {
...
+						Source: "registry.example/missing:1.0",
+						Target: "/test_mount",
+					},
+				},
+			},
+		},
+	}
+
+	_, err := p.WithImagesResolved(resolver)
+	assert.Error(t, err, "manifest unknown")
```
glours:
```suggestion
	assert.ErrorContains(t, err, "manifest unknown")
```

glours (changes requested):
Thanks for the iteration — memoization, the four additional test cases, and the updated doc comment all address my earlier feedback, and the core change is sound (deep-copy isolation holds, concurrency is safe, single-image behavior is preserved).    
                                                                                                                                                                                                                                                             
  A few targeted suggestions left inline:                                                                                                                                                                                                                    
  - one test that conflates canonical detection with the cache and could mask a future regression,                                                                                                                                                           
  - the per-service cache scope (a project-wide cache would dedupe across services too),                                                                                                                                                                     
  - a doc-comment phrasing that overstates the dedup guarantee (could be fix by the per-project cache above ☝️ )                                                                                                                                                                                              
  - plus two small nits on a parameter name and a brittle error assertion.

ssam18:
Thanks @glours! All points addressed in the latest push, and the branch is rebased on current main:

- **Project-wide cache**: hoisted to `WithImagesResolved` as a `sync.Map` as suggested. Since `WithServicesTransform` runs services concurrently, I also wrapped lookups in `golang.org/x/sync/singleflight` (already a dependency via errgroup)  a bare `sync.Map` would still let two services racing on a cache miss each hit the resolver, so singleflight is what actually guarantees one resolver call per distinct reference. It also makes the new cross-service dedup test deterministic.
- **Doc comment**: reworded to describe the project-wide dedup guarantee.
- **alreadyCanonical test**: dropped `service.Image` so only the image-mount path is exercised and `resolverCalls == 0` can only come from canonical detection.
- **New test**: `Test_ResolveImages_imageMount_sharedAcrossServices` three services sharing one tag, asserting a single resolver call (race-clean under `-race -count=20`).
- **Nits**: dropped the `named` parameter name in the resolver type literal, and switched to `assert.ErrorContains`.

ndeloof (approved):

ndeloof:
fixed by https://github.com/compose-spec/compose-go/pull/894

PR #886, opened 2026-06-19
loader: memoize includes to avoid re-expanding diamond include graphs

davireis:
## Problem

`ApplyInclude` re-parses and recursively re-expands an included file **once per include path that reaches it**. When the same file is reachable through more than one path (a "diamond" in the include graph), this is quadratic-to-exponential: a 24-level doubling graph loads the leaf 2²⁴ ≈ 16.7M times.

This shows up in real monorepos that aggregate per-target / per-project compose fragments via `include:`. A federation of ~80 services took ~55s in `docker compose config`; the cost is re-expansion, not the graph size.

## Fix

Memoize each loaded include model for the duration of a single load (carried in `ctx`, so it never leaks across `Load` calls). The cache key is **every input that determines the model** — resolved paths, working dir, project dir, and effective environment — and a deep copy is handed out on each hit.

The merge into the parent (`importResources`) still runs for **every** occurrence, so:
- a same-file `extends` in the including file still resolves (the included content is present in each including scope), and
- the result is byte-identical to loading the file each time; only the parse + recursive expansion is shared.

### Correctness details
- **Per-include `env_file` / `project_directory`**: folded into the key, so the same path included with a different environment or project dir does not share a cache entry.
- **Relative-path rebasing**: the same file reached through two parents can have a different relative base (e.g. `a/b` vs `b`), which yields models with different relative paths. Keying on the working dir prevents reusing a model whose paths the caller would then rebase incorrectly.
- **Cycle-safe**: an include cycle is intrinsic to a node's subtree (the back-edge is in the fixed set of files the node includes), so it is detected on the node's *first* load — before it can be cached. A cyclic node is therefore never served from cache.
- **Listeners** still fire per include entry; only the load is memoized.

## Tests
- `TestIncludeDiamondDedup`: a depth-24 diamond that **times out without the cache** and completes in ~0.15s with it (both a correctness and a non-flaky perf-regression test).
- `BenchmarkIncludeDiamond`.
- Full `go test ./...` passes; `gofmt`/`go vet` clean.

### DIFF
--- a/loader/include.go
+++ b/loader/include.go
@@ -50,9 +50,31 @@ import (
 // Cycle-safe: an include cycle is intrinsic to a node's subtree (the back-edge
 // is in the fixed set of files the node includes), so it is detected on the
 // node's first load, which fails before the node can be cached.
+//
+// Listener fidelity: loadYamlModel emits listener events (notably "extends", and
+// nested "include") for the subtree it expands. Memoizing that call would drop
+// those events on every cache hit, silently changing the public Listener
+// contract (e.g. an extends-counting consumer would see a count that depends on
+// include topology). Each entry therefore records the events emitted on first
+// load and replays them, in order, on every cache hit — see ApplyInclude.
 type includeCache struct {
 	mu      sync.Mutex
-	entries map[string]map[string]any
+	entries map[string]includeCacheEntry
+}
+
+// includeCacheEntry is a memoized include: the expanded model plus the listener
+// events emitted while expanding it, replayed on cache hits to keep per-include
+// listeners firing once per occurrence.
+type includeCacheEntry struct {
+	model  map[string]any
+	events []recordedEvent
+}
+
+// recordedEvent is a single listener event captured during an include load so it
+// can be replayed verbatim on a later cache hit.
+type recordedEvent struct {
+	event    string
+	metadata map[string]any
 }
 
 type includeCacheKey struct{}
@@ -64,23 +86,23 @@ func getOrCreateIncludeCache(ctx context.Context) (*includeCache, context.Contex
 	if c, ok := ctx.Value(includeCacheKey{}).(*includeCache); ok {
 		return c, ctx
 	}
-	c := &includeCache{entries: map[string]map[string]any{}}
+	c := &includeCache{entries: map[string]includeCacheEntry{}}
 	return c, context.WithValue(ctx, includeCacheKey{}, c)
 }
 
-func (c *includeCache) get(key string) (map[string]any, bool) {
+func (c *includeCache) get(key string) (includeCacheEntry, bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if m, ok := c.entries[key]; ok {
-		return deepCopyMapping(m), true
+	if e, ok := c.entries[key]; ok {
+		return includeCacheEntry{model: deepCopyMapping(e.model), events: e.events}, true
 	}
-	return nil, false
+	return includeCacheEntry{}, false
 }
 
-func (c *includeCache) put(key string, model map[string]any) {
+func (c *includeCache) put(key string, model map[string]any, events []recordedEvent) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.entries[key] = deepCopyMapping(model)
+	c.entries[key] = includeCacheEntry{model: deepCopyMapping(model), events: events}
 }
 
 // includeKey hashes the inputs that determine an included model. Two include
@@ -92,27 +114,41 @@ func (c *includeCache) put(key string, model map[string]any) {
 // a different relative base (e.g. "a/b" vs "b"), which yields models with
 // different relative paths. Keying on it avoids reusing a model whose paths the
 // caller would then rebase incorrectly.
+//
+// Encoding: every field is length-prefixed and each variable-length section is
+// count-prefixed. A bare separator byte is not safe here — types.Mapping is
+// populated from .env files / process env, where any byte (NUL included) is a
+// legal key or value, so a value could otherwise impersonate a separator and
+// two distinct (paths, env) tuples could hash to the same key, serving a wrong
+// cached model with no error surfaced. Length/count prefixes make the byte
+// stream uniquely decodable, so distinct inputs always produce distinct keys.
+//
+// Note: Substitute and TypeCastMapping are intentionally excluded from the key.
+// They are invariant across includes within a single Load (cloned unchanged from
+// the top-level options at the call site). If a future option lets them vary per
+// include, they must be folded into the key.
 func includeKey(paths []string, workingDir, projectDir string, env types.Mapping) string {
 	h := sha256.New()
-	for _, p := range paths {
-		_, _ = h.Write([]byte(p))
-		_, _ = h.Write([]byte{0})
+	writeInt := func(n int) { _, _ = fmt.Fprintf(h, "%d:", n) }
+	write := func(s string) {
+		_, _ = fmt.Fprintf(h, "%d:", len(s))
+		_, _ = h.Write([]byte(s))
 	}
-	_, _ = h.Write([]byte{1})
-	_, _ = h.Write([]byte(workingDir))
-	_, _ = h.Write([]byte{1})
-	_, _ = h.Write([]byte(projectDir))
-	_, _ = h.Write([]byte{1})
+	writeInt(len(paths))
+	for _, p := range paths {
+		write(p)
+	}
+	write(workingDir)
+	write(projectDir)
 	keys := make([]string, 0, len(env))
 	for k := range env {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
+	writeInt(len(keys))
 	for _, k := range keys {
-		_, _ = h.Write([]byte(k))
-		_, _ = h.Write([]byte{0})
-		_, _ = h.Write([]byte(env[k]))
-		_, _ = h.Write([]byte{0})
+		write(k)
+		write(env[k])
 	}
 	return hex.EncodeToString(h.Sum(nil))
 }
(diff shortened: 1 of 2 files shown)


loader/include.go
```diff
@@ -30,6 +34,114 @@ import (
...
+	for k := range env {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		_, _ = h.Write([]byte(k))
+		_, _ = h.Write([]byte{0})
+		_, _ = h.Write([]byte(env[k]))
+		_, _ = h.Write([]byte{0})
+	}
```
glours, line 127: The single-byte separators written here (`[]byte{0}` on lines 99, 113, 115 and `[]byte{1}` on lines 101, 103, 105) aren't unambiguous against env values that contain those bytes. `types.Mapping` is `map[string]string` populated from `.env` files /
  process env, where embedded NULs aren't forbidden. Two distinct `(paths, env)` tuples can then produce the same byte stream and hash to the same key, wrong cached model served, no error surfaced.

Toy example: `env = {"A\x00B": "X"}` and `env = {"A": "B\x00X"}` both serialize to `…A\x00B\x00X\x00` through the loop on lines 111-116.

Length-prefixing each field removes the ambiguity:

  ```go
  write := func(s string) {
      fmt.Fprintf(h, "%d:", len(s))
      _, _ = h.Write([]byte(s))
  }
  for _, p := range paths { write(p) }
  write(workingDir)
  write(projectDir)
  for _, k := range keys { write(k); write(env[k]) }
```
While you're here, a short comment explaining why Substitute / TypeCastMapping are intentionally excluded from the key (invariant across includes within a single load) would help a future contributor who adds a per-subtree option not silently introduce collisions, like 

```go
// Note: Substitute and TypeCastMapping are intentionally excluded from the key.
// They are invariant across includes within a single Load (cloned unchanged from
// the top-level options at the call site above). If a future option allows them
// to vary per include, they must be folded into the key.
```

loader/include.go
```diff
@@ -151,9 +265,19 @@ func ApplyInclude(ctx context.Context, workingDir string, environment types.Mapp
...
+		// handed out), so any same-file `extends` in the including file still
+		// resolves and the result is identical to loading it each time.
+		key := includeKey(r.Path, config.WorkingDir, r.ProjectDirectory, config.Environment)
+		imported, ok := cache.get(key)
+		if !ok {
+			imported, err = loadYamlModel(ctx, config, loadOptions, &cycleTracker{}, included)
+			if err != nil {
+				return err
+			}
+			cache.put(key, imported)
```
glours: There is a silent change to the public `Listener` contract here.
 
When `cache.get(key)` returns `ok` at line 274, the `if !ok` block (lines 275-281) is skipped, so `loadYamlModel(ctx, config, loadOptions, &cycleTracker{}, included)` on line 276 does not run on cache hits. The `extends` events emitted from inside
  that call path, see `opts.ProcessEvent("extends", v)` at `loader/extends.go:76` and `opts.ProcessEvent("extends", map[string]any{"service": ref})` at `extends.go:79`, are therefore not re-emitted for any subsequent diamond-traversal of the same included file. Compare with the `include` listener at `include.go:175`, which fires per occurrence (it's emitted in the outer loop, before the cache lookup): two events documented in the same pipeline now behave asymmetrically.

That isn't a cosmetic observability change:

  - `Listener` is part of the public API (`loader/loader.go:109`), and downstream tools wire into it via `cli/options.go:83`.
  - The existing test `TestLoadExtendsListener` at `loader/extends_test.go:440` asserts `assert.Equal(t, extendsCount, 3)`, which establishes the contract: one event per resolved `extends`. Diamond include topologies break that invariant after this PR,
   and silently, because no test in this repo combines diamond includes with extends listeners.
  - Any consumer counting extends events for telemetry, dependency tracking, audit, or progress reporting will see different counts depending on the include topology , exactly the kind of regression that surfaces months later as a confused user report.

I'd suggest preserving the contract by recording events on first load and replaying them on cache hit.

loader/include_test.go
```diff
@@ -243,3 +244,47 @@ func createFileSubDir(t *testing.T, rootDir, subDir, content, fileName string) {
...
+	assert.NilError(t, os.WriteFile(filepath.Join(dir, fmt.Sprintf("level%d.yaml", depth)), []byte(leaf), 0o600))
+
+	p, err := LoadWithContext(context.TODO(), types.ConfigDetails{
+		WorkingDir:  dir,
+		ConfigFiles: []types.ConfigFile{{Filename: filepath.Join(dir, "level0.yaml")}},
+	}, withProjectName("diamond", true))
+	assert.NilError(t, err)
+	_, err = p.GetService("leaf")
+	assert.NilError(t, err)
+}
```
glours, line 363: Nice regression test, but the only assertions are `assert.NilError(t, err)` on line 267 and the `GetService("leaf")` check on lines 268-269. If the cache regresses, the `LoadWithContext` call on line 263 will spin on `2^24` leaf loads and the test goroutine stays parked on that line — `loader/` doesn't check `ctx.Done()` anywhere, so a `context.WithTimeout` wouldn't help. The test then hangs until the default `go test` timeout (10 minutes), and fails with a generic timeout message, slow CI feedback, unclear failure mode.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                             
A fast, descriptive fail requires running the load in a goroutine and selecting on a timeout. The leaked goroutine on timeout is fine, `t.Fatal` ends the test and the process exits soon after.                                                          
   
  ```go           
  type result struct {
      p   *types.Project
      err error
  }

  // LoaderWithContext don't properly support context today, so the timeout below is consumed by the
  // select, not by LoadWithContext. Two separate contexts keep that intent clear.                                                                                                                                                                           
  timeout, cancel := context.WithTimeout(t.Context(), 5*time.Second)                                                                                                                                                                                         
  defer cancel()                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                             
  done := make(chan result, 1)
  go func() {                                                                                                                                                                                                                                                
      p, err := LoadWithContext(t.Context(), types.ConfigDetails{
          WorkingDir:  dir,                                                                                                                                                                                                                                  
          ConfigFiles: []types.ConfigFile{{Filename: filepath.Join(dir, "level0.yaml")}},
      }, withProjectName("diamond", true))                                                                                                                                                                                                                   
      done <- result{p, err}
  }()                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                             
  select {
  case r := <-done:                                                                                                                                                                                                                                          
      assert.NilError(t, r.err)
      _, err := r.p.GetService("leaf")
      assert.NilError(t, err)
  case <-timeout.Done():
      t.Fatal("diamond include did not complete within 5s — cache likely not working")                                                                                                                                                                       
  }
```
                                                                                                                                                                                                                                                             
  5s is generous; the current run completes in ~0.18s on my machine.

**Review by glours — CHANGES_REQUESTED** [review]
Thanks for the contribution. Three inline points to address.                                                                                                                                                                                               
                                                                                                                                                                                                                                                             
  - A cache-key separator collision that's fixable with length-prefixed hashing.                                                                                                                                                                             
  - **More importantly**, the `extends` listener events emitted inside `loadYamlModel` are silently dropped on cache hits, which changes the public `Listener` contract (cf. `TestLoadExtendsListener` asserting an exact count). Worth treating as a blocker                                                                                                                                                       
  - A time-bound guard on the diamond test, so a future cache regression fails fast with a clear message instead of hanging CI for the default 10-minute timeout.
                                                                                                                                                                                                                                                             
The overall approach (per-load cache, env+paths+workingDir+projectDir as key, deep-copy on get) is the right shape — cycle safety, deep-copy completeness, and the unexported context key all check out.

File: loader/include.go (hunk 1 above)
Comment (davireis) [reply loader/include.go:127]: Fixed in 52aa7ae. `includeKey` now length-prefixes every field and count-prefixes the path list and env map, so the byte stream is uniquely decodable and no env value can impersonate a separator. I added the count prefixes (beyond plain length-prefixing) to also close a structural boundary collision — e.g. `paths=["p","q"], wd="w", pd="d", env={}` vs `paths=[], wd="p", pd="q", env={"w":"d"}` would otherwise serialize identically.

Also added the comment on why `Substitute`/`TypeCastMapping` are excluded, and a `TestIncludeKeyNoCollision` covering your NUL toy example plus the structural cases.

File: loader/include.go (hunk 2 above)
Comment (davireis) [reply loader/include.go]: Fixed in 52aa7ae by recording the events emitted on the first load and replaying them on cache hits, as you suggested — so a hit is indistinguishable from a fresh load to a listener, and `extends` (and nested `include`) events fire once per occurrence again.

One wrinkle worth flagging: replaying unconditionally makes the *event stream* itself exponential on a doubling graph (each hit replays its subtree’s events back through the ancestor recorders), even though model expansion stays memoized — `TestIncludeDiamondDedup` went from ~0.15s to ~340s. Since there is no contract to preserve when nobody is listening, recording is skipped entirely when no listeners are registered, which keeps the common path cheap; with listeners attached the count faithfully matches an uncached load.

Added `TestIncludeDiamondListener`: a diamond over a file with one `extends`, asserting the event fires twice (once per path). It asserts 1 with the replay removed and 2 with it.

File: loader/include_test.go (hunk 3 above)
Comment (davireis) [reply loader/include_test.go:363]: Done in 52aa7ae. `TestIncludeDiamondDedup` now runs the load in a goroutine and selects on a separate timer, so a cache regression fails fast with a clear message instead of parking on `LoadWithContext` until the 10-minute `go test` default. I kept the two-context split and the comment noting the loader does not observe `ctx.Done()` today, so the timeout is consumed by the select rather than by the load.

I used a 30s timer rather than 5s purely for headroom under `-race`/loaded CI; the load completes in ~0.14s here, so it still fails fast against any real regression.

**Comment by davireis** [comment]
Thanks for the thorough review @glours — all three points addressed in 52aa7ae and pushed:

- **Cache-key collision:** length-prefixed + count-prefixed hashing in `includeKey`, the `Substitute`/`TypeCastMapping` exclusion comment, and a new `TestIncludeKeyNoCollision`.
- **Listener contract (blocker):** events are recorded on first load and replayed on cache hits so `extends`/nested `include` fire once per occurrence; recording is skipped when no listeners are registered to avoid re-introducing exponential event replay. New `TestIncludeDiamondListener` pins the per-occurrence count.
- **Diamond test guard:** the load runs in a goroutine with a select-on-timer, failing fast with a clear message instead of hanging until the `go test` default.

`go test ./...` (all packages) and `go test -race ./loader/...` pass; `gofmt`/`go vet` clean. Replies inline on each thread.

**Comment by glours** [comment]
@davireis you need to sign-off your commits for legal purpose, please

**Comment by bonitao** [comment]
> @davireis you need to sign-off your commits for legal purpose, please

Done.

File: loader/include.go [hunk 4]
```diff
@@ -151,11 +276,45 @@ func ApplyInclude(ctx context.Context, workingDir string, environment types.Mapp
...
+		entry, ok := cache.get(key)
+		switch {
+		case ok:
+			// Replay so per-occurrence listeners (extends, nested include) fire
+			// on this traversal too: a cache hit must look identical to a fresh
+			// load.
+			for _, ev := range entry.events {
+				options.ProcessEvent(ev.event, deepCopyMapping(ev.metadata))
+			}
+		case len(loadOptions.Listeners) == 0:
```
ndeloof, line 293: inside Docker Compose, `loadOptions.Listeners` is set - so this "fix" will never apply 😢
Also need to mention the fix here is set to preserve loading speed, but will increase memory usage as loaded includes are kept in cache until Load completion. Depending on the context, this _may_ not be the expected behavior: consider a deployment automation which is ran in a memory-constrained environment. After a compose update it is in charge to refresh the stack, being bit slow is not nice, but not a blocker. With this fix it could be killed by OOM

PR #890, opened 2026-06-29
perf(schema): compile compose-spec schema once

ndeloof:
## What

`schema.Validate` rebuilt a `jsonschema.Compiler` and recompiled the embedded `compose-spec.json` schema — including loading and resolving the draft 2020-12 meta-schema — on **every** call. This commit compiles the schema **once** behind a `sync.Once` and reuses the resulting `*jsonschema.Schema`.

Also switches `MustCompile` → `Compile` so a compilation failure is returned as an error instead of panicking, and adds a concurrent `Validate` regression test (run with `-race`).

## Why

The loader calls `Validate` once per file on the hot path, so recompiling the full schema each time is needlessly expensive. Compiling once:
- keeps `Validate` cheap (~µs/op instead of a full schema compilation per call);
- makes `Validate` safe to call concurrently without sharing any mutable compiler state.

## Context

Investigated while looking at docker/compose#13866 (`fatal error: concurrent map iteration and map write` reported inside schema compilation). Note: I could **not** reproduce that crash through `schema.Validate` — even under heavy concurrency with `-race`, both before and after this change, no data race fires. The current loader is sequential and each `Validate` call already keeps independent state, so the reported concurrency most likely originates in the caller (docker/compose). This change is therefore offered as a **performance + hardening** improvement rather than a fix for that specific crash; it relates to the issue without claiming to close it.

## Test

- `go test -race ./schema/...` passes, including the new `TestValidateConcurrent` (50 goroutines) and an ad-hoc hammer of 4000 concurrent validations of the real schema.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/santhosh-tekuri/jsonschema/v6"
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

PR #895, opened 2026-07-13
interpolation: sort map keys for deterministic error reporting

mahesh-sadupalli:
## Problem

`docker compose config --quiet` (and `docker compose up`) can report a different
missing-required-variable error on each run when multiple config sections reference
absent required variables. For example, one run may report:

```
error while interpolating services.traefik.environment.[]: required variable TIMEZONE is missing a value
```

and the next may report:

```
error while interpolating services.traefik.volumes.[].source: required variable TRAEFIK_ACME_PATH is missing a value
```

Root cause: `recursiveInterpolate` and `Interpolate` iterate over
`map[string]interface{}` directly. Go randomises map iteration order, so the
first missing variable encountered varies between runs.

## Fix

Sort the map keys alphabetically before iterating in both `Interpolate` and
`recursiveInterpolate`. This adds a small `sortedKeys` helper and changes two
`range` loops to range over a sorted key slice instead.

The output `map[string]interface{}` is unchanged — only the order in which errors
are encountered becomes deterministic. With alphabetical traversal the error now
always identifies the lexicographically first problematic path, giving users
a stable, reproducible message to act on.

## Test

A new `TestInterpolateRequiredErrorIsDeterministic` test invokes `Interpolate`
100 times on a config with two map keys (`"aaa"` and `"zzz"`) both referencing
a missing required variable, and asserts that every run reports the same error
and that the path contains `"aaa"` (the alphabetically first key).

Fixes docker/compose#13712

### DIFF
--- a/interpolation/interpolation.go
+++ b/interpolation/interpolation.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 
 	"github.com/compose-spec/compose-go/v2/template"
 	"github.com/compose-spec/compose-go/v2/tree"
@@ -58,8 +59,8 @@ func Interpolate(config map[string]interface{}, opts Options) (map[string]interf
 
 	out := map[string]interface{}{}
 
-	for key, value := range config {
-		interpolatedValue, err := recursiveInterpolate(value, tree.NewPath(key), opts)
+	for _, key := range sortedKeys(config) {
+		interpolatedValue, err := recursiveInterpolate(config[key], tree.NewPath(key), opts)
 		if err != nil {
 			return out, err
 		}
@@ -88,8 +89,8 @@ func recursiveInterpolate(value interface{}, path tree.Path, opts Options) (inte
 
 	case map[string]interface{}:
 		out := map[string]interface{}{}
-		for key, elem := range value {
-			interpolatedElem, err := recursiveInterpolate(elem, path.Next(key), opts)
+		for _, key := range sortedKeys(value) {
+			interpolatedElem, err := recursiveInterpolate(value[key], path.Next(key), opts)
 			if err != nil {
 				return nil, err
 			}
@@ -125,6 +126,15 @@ func newPathError(path tree.Path, err error) error {
 	default:
 		return fmt.Errorf("error while interpolating %s: %w", path, err)
 	}
+}
+
+func sortedKeys(m map[string]interface{}) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
 }
 
 func (o Options) getCasterForPath(path tree.Path) (Cast, bool) {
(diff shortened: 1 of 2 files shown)


interpolation/interpolation.go
```diff
@@ -127,6 +128,15 @@ func newPathError(path tree.Path, err error) error {
...
 
+func sortedKeys(m map[string]interface{}) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
```
glours, line 139: This helper already exists in the repo: `utils.MapKeys` (`utils/collectionutils.go`) does exactly this, generically, and is already used for the same "iterate map keys deterministically" purpose in `types/project.go` and `graph/cycle.go`.                                                                                                                                                                                                                       
                                                                                                                                                                                                                                          
Could you drop `sortedKeys` (and the `sort` import) and use it instead?

interpolation/interpolation_test.go
```diff
@@ -223,6 +224,34 @@ func TestInterpolateWithCast(t *testing.T) {
...
+	config := map[string]interface{}{
+		"service": map[string]interface{}{
+			"zzz": "${MISSING:?required}",
+			"aaa": "${MISSING:?required}",
+		},
+	}
+
+	var firstErr string
+	for range 100 {
+		_, err := Interpolate(config, Options{})
```
glours, line 239: `Options{}` makes `LookupValue` fall back to `os.LookupEnv`, so the test silently depends on `MISSING` not being set in the process environment. Passing `Options{LookupValue: defaultMapping}` like the other tests in this file makes it hermetic.

interpolation/interpolation_test.go
```diff
@@ -223,6 +224,34 @@ func TestInterpolateWithCast(t *testing.T) {
...
+		if firstErr == "" {
+			firstErr = err.Error()
+		} else if err.Error() != firstErr {
+			t.Fatalf("non-deterministic error: got %q, want %q", err.Error(), firstErr)
+		}
+	}
+	// Alphabetically first key ("aaa") must appear in the path.
+	if !strings.Contains(firstErr, "service.aaa") {
+		t.Fatalf("expected error path to contain 'service.aaa', got: %s", firstErr)
+	}
```
glours, line 252: We can use the assert library for this. Checking the exact error message would also make the second assertion unnecessary, since it would already verify that the alphabetical ordering is correct.

```suggestion
	for range 100 {                                                                                                                                                                                                                   
                _, err := Interpolate(config, Options{LookupValue: defaultMapping})                                                                                                                                                       
                assert.Error(t, err, "error while interpolating service.aaa: required variable MISSING is missing a value: required")                                                                                                     
        }          
```

glours (changes requested):
I don’t think this PR addresses the root cause. Instead of returning errors one at a time, we should return all of them at once. I’ve opened a separate PR that takes this approach 
https://github.com/compose-spec/compose-go/pull/903

mahesh-sadupalli:
Thanks for the review and the improved approach in #903 — collecting all errors is clearly the right fix. Closing this in favour of #903.

PR #896, opened 2026-07-17
loader: apply service selection to the raw yaml model path

glours:
LoadModelWithContext ignored Options.SelectedServices: selection only existed on the typed path (ModelToProject), so consumers loading the raw model — like docker compose config --no-interpolate <SERVICE> — always got every service back.

SelectModelServices restricts the dict to the selected services plus their transitive depends_on closure. Unresolved dependency keys, such as an uninterpolated ${VAR}, are skipped rather than failing the load, since the raw model may be only partially resolved. This completes for the model path the selection support #875 added to the typed path.

Addresses docker/compose#13614


glours:
We ultimately decided with @ndeloof not to handle depends_on filtering in the raw model list. Instead, we’ll handle it on the Compose side by showing a warning that --no-interpolate does not respect service filtering.

PR #900, opened 2026-07-24
Add `!merge` tag for positional sequence merge

ndeloof:
> [!NOTE]
> Prototype / draft for discussion.

## What

Add a `!merge` YAML tag that makes an override sequence merge **positionally**
(element by element, by index) instead of the default *append*.

Today sequences can only be appended (default / `mergeToSequence`) or fully
replaced (`!override`, or the `override` rule for `command`/`entrypoint`/…).
There is no way to change a single element of a list in place. `!merge` fills
that gap:

```yaml
# compose.yaml
services:
  app:
    command: [server, --port, "8080", --verbose]

# compose.override.yaml
services:
  app:
    command: !merge
      - {}        # keep base[0]  -> server
      - {}        # keep base[1]  -> --port
      - "9090"    # replace base[2]
                  # base[3] --verbose kept (override shorter)
# result: [server, --port, "9090", --verbose]
```

Semantics, per index `i`:

- a **no-op** element (`- {}` or `- null`) keeps `base[i]`;
- otherwise `override[i]` is **deep-merged** with `base[i]` (deep-merge for
  mappings, replace for scalars);
- override longer → the sequence is extended; base longer → the tail is kept.

The tag is **ignored on anything that is not a sequence**, and the default
behavior is **unchanged** when the tag is absent (opt-in, non-breaking). `!merge`
takes precedence over both the default append and the per-field merge rules,
since it is an explicit user request.

## How

Mirrors the existing `!reset`/`!override` machinery:

- `loader/reset.go` — `ResetProcessor` records the paths of `!merge`-tagged
  sequences and strips the tag so the node decodes as a plain sequence.
- `loader/loader.go` — passes those paths to the merge step.
- `override/merge.go` — new `MergeWithPositionalPaths` / `mergePositional`
  performs the positional merge at the recorded paths. `Merge`/`MergeYaml` stay
  as-is (wrappers), so the public API is unchanged.

Tests double as documentation: `override/merge_positional_test.go` (semantics)
and `loader/merge_positional_test.go` (the tag end to end).

## Known limitations (prototype)

- Positional paths are not threaded through the special mergers (`build`,
  `networks`, `depends_on`, …), so a `!merge` nested *under* one of those fields
  is not honored yet.
- A `!merge` sequence reused through a YAML anchor is not cached/replayed at the
  reuse site.

Both are addressable follow-ups if the approach is accepted.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -497,7 +497,11 @@ func loadYamlFile(ctx context.Context,
 
 		}
 
-		dict, err = override.Merge(dict, cfg)
+		var mergePaths []tree.Path
+		if rp, ok := processor.(*ResetProcessor); ok {
+			mergePaths = rp.MergePaths()
+		}
+		dict, err = override.MergeWithPositionalPaths(dict, cfg, mergePaths)
 		if err != nil {
 			return err
 		}
--- a/loader/merge_positional_test.go
+++ b/loader/merge_positional_test.go
@@ -1,0 +1,113 @@
+/*
+   Copyright 2020 The Compose Specification Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package loader
+
+import (
+	"context"
+	"testing"
+
+	"github.com/compose-spec/compose-go/v2/types"
+	"gotest.tools/v3/assert"
+)
+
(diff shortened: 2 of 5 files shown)


glours:
My take: I don't think we should merge this as designed.

I couldn't find any real user need behind it, no issue, no complaint, nothing asking for this. And the motivating example (changing one arg in `command`) is already solvable today since `command`/`entrypoint` fully replace on override anyway, so the actual gap this closes looks very thin.

On the "known limitations" you listed, the one about `!merge` not being threaded through the special mergers (`build`, `networks`, `depends_on`, etc.)
I don't see that as an addressable follow-up, I'd treat it as blocking as-is. It's not just unsupported, it's silently ignored: falls back to plain append with no error. 
For a library that feeds config straight into the Compose CLI, that's worse than failing loudly, someone writes `!merge`, thinks it worked, and it didn't.

The implementation also feels like it's fighting the existing merge code rather than fitting into it — a new parameter gets threaded through pretty much every merge function just to reach the one place that needs it. 

If we do want this kind of capability, I'd rather scope it to a specific, curated list of attributes (the same way we already special-case `command`/`entrypoint` for replace) instead of a generic tag that applies to any sequence anywhere and overrides every other merge rule. Much smaller blast radius for the same benefit.

WDYT?

PR #903, opened 2026-07-29
fix(interpolation): report all errors in a deterministic order

glours:
Interpolation used to stop at the first error, and Go's random map iteration made a different missing required variable surface on each run. 
Collect every interpolation error while walking the whole config instead, and sort the collected errors by message before joining them, so both the reported set and its order are stable across runs and users can fix all missing variables at once. 
Sorting only happens on the error path, which keeps the successful-interpolation hot path allocation-identical to the previous behavior.

Fixes docker/compose#13712

### DIFF
--- a/interpolation/interpolation.go
+++ b/interpolation/interpolation.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 
 	"github.com/compose-spec/compose-go/v2/template"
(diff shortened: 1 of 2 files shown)


ndeloof (approved):
LGTM, with a proposal:
as interpolation always uses `newPathError` to produce error, we could use a custom error type which knows the path in error, and allows sorting by key (vs sorting by error string)

ndeloof (approved):

PR #907, opened 2026-08-07
fix(override): merge ulimits into the base value

vsaraikin:
Set one ulimit key in the base file and a different one in the override, and the base key disappears:

```yaml
# base                # override
ulimits:              ulimits:
  locks:                locks:
    soft: 20000           hard: 65535
    hard: 40000

# got                 # want
locks:                locks:
  hard: 65535           soft: 20000
                        hard: 65535
```

`mergeUlimit` receives the base value and the override value, but it reads both mappings out of the override:

```go
func mergeUlimit(_ any, o any, p tree.Path) (any, error) {
	over, ismapping := o.(map[string]any)
	if base, ok := o.(map[string]any); ok && ismapping {
		return mergeMappings(base, over, p)
```

So it merges the override with itself. The base argument is named `_` and never read. Every other function in `mergeSpecials` takes its first argument as the base, this one is the exception.

`Test_mergeYamlUlimits` already covered the case, but its expectation encoded the old behaviour, so it is updated to expect the base `soft` to survive. Reverting `merge.go` alone makes it fail. Full suite is green

### DIFF
--- a/override/merge.go
+++ b/override/merge.go
@@ -224,9 +224,9 @@ func convertIntoSequence(value any) []any {
 	return nil
 }
 
-func mergeUlimit(_ any, o any, p tree.Path) (any, error) {
+func mergeUlimit(c any, o any, p tree.Path) (any, error) {
 	over, ismapping := o.(map[string]any)
-	if base, ok := o.(map[string]any); ok && ismapping {
+	if base, ok := c.(map[string]any); ok && ismapping {
 		return mergeMappings(base, over, p)
 	}
 	return o, nil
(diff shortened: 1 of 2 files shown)


ndeloof (approved):

ndeloof (approved):
Verified locally: the updated test fails with the previous merge.go (the old expectation encoded the buggy output) and the override+loader suites are green with the fix. Merge semantics for the scalar/mapping mismatch cases are unchanged. I'll follow up with an additional test covering short/long syntax combinations.

PR #909, opened 2026-08-10
cli: validate COMPOSE_FILE entries point to actual compose files

ndeloof:
Entries set by the `COMPOSE_FILE` environment variable used to be resolved to absolute paths with a bare `os.Stat`, so an empty or directory entry made the loader fail later with a cryptic OS-level error such as `read <pwd>: is a directory` (an empty entry resolves to the working directory — see docker/compose#13649).

Each entry is now validated when selected: it must be stdin (`-`), a remote resource supported by a registered `ResourceLoader`, or a regular local file. Errors state how the file was selected:

```
compose file "" set by COMPOSE_FILE environment variable is invalid: /home/user is a directory
```

Implicitly selected config paths (`COMPOSE_FILE`, default file discovery) are tracked so read errors get the same context, while explicitly selected files (`-f`) keep a plain `compose file X is invalid` prefix.

Side effect: remote resource references in `COMPOSE_FILE` are no longer rejected by the local `os.Stat` check (they still require an explicit `COMPOSE_PATH_SEPARATOR`, as the default `:` conflicts with remote references).

---
*This contribution was prepared by an AI agent acting on the maintainer's behalf, then reviewed by them.*

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -18,6 +18,7 @@ package cli
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

PR #911, opened 2026-08-11
fix: merge attributes of repeated variable occurrences in ExtractVariables

### LINKED ISSUE #13718: [BUG] env variable detected as not required even though it is
### Description

Given a service with the following labels:

```yaml
services:
  pihole:
    image: pihole/pihole:2026.02.0
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.pihole.rule=Host(`${PIHOLE_DOMAIN:?}`)"
      - "traefik.http.services.pihole.loadbalancer.server.port=80"
      # Always redirect / to /admin/
      - "traefik.http.middlewares.pihole-redirect.redirectregex.regex=^(https://${PIHOLE_DOMAIN})/?$"
```

`docker compose config --variables --format json` shows `PIHOLE_DOMAIN` as `Required=false` when it is required

### Steps To Reproduce

With the above config, run:

```shell
docker compose config --variables --format json
```

Output:

```json
{
    "PIHOLE_DOMAIN": {
        "Name": "PIHOLE_DOMAIN",
        "DefaultValue": "",
        "PresenceValue": "",
        "Required": false
    }
}
```

Also run:

```shell
docker compose config --quiet
```

Output:

```console
$ docker compose config --quiet
error while interpolating services.pihole.labels.[]: required variable PIHOLE_DOMAIN is missing a value
```

### Compose Version

```Text
Docker Compose version v2.40.3
```

### Docker Environment

```Text

```

### Anything else?

_No response_

ndeloof:
`ExtractVariables` merged repeated occurrences of the same variable by plain overwrite: whichever occurrence happened to be processed last won, and map iteration made the outcome non-deterministic. A plain `${VAR}` occurrence could erase the `Required` flag extracted from a `${VAR:?}` one, so `docker compose config --variables` reported `Required: false` for a variable that interpolation actually requires (docker/compose#13718).

Merge occurrences instead:
- a variable is required as soon as one of its occurrences is — interpolation of that occurrence fails when the variable is unset, whatever the other occurrences declare;
- the first non-empty default/presence value is kept, so a plain occurrence doesn't erase them.

Verified end-to-end against docker/compose: `config --quiet` fails with "required variable PIHOLE_DOMAIN is missing a value" while `config --variables` used to report `Required: false`; with this change it reports `Required: true`.

Fixes docker/compose#13718

### DIFF
--- a/template/variables.go
+++ b/template/variables.go
@@ -44,27 +44,50 @@ func recurseExtract(value interface{}, pattern *regexp.Regexp) map[string]Variab
 	case string:
 		if values, is := extractVariable(value, pattern); is {
 			for _, v := range values {
-				m[v.Name] = v
+				combineVariable(m, v)
 			}
 		}
 	case map[string]interface{}:
 		for _, elem := range value {
 			submap := recurseExtract(elem, pattern)
-			for key, value := range submap {
-				m[key] = value
+			for _, value := range submap {
+				combineVariable(m, value)
 			}
 		}
 
 	case []interface{}:
 		for _, elem := range value {
 			submap := recurseExtract(elem, pattern)
-			for key, value := range submap {
-				m[key] = value
+			for _, value := range submap {
+				combineVariable(m, value)
 			}
 		}
 	}
 
 	return m
+}
+
+// combineVariable merges a new occurrence of a variable into m: a variable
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

PR #914, opened 2026-08-17
tests: codify the testing contract

ndeloof:
compose-go's tests are the executable counterpart of the Compose Specification — and, increasingly, the surface maintainers and their coding agents communicate through. Five dedicated commits codify that contract:

- **TESTING.md** elevates the two patterns that emerged in this codebase (attribute conformance files in `loader/tests/`, YAML-in/YAML-out expectations in `override/`) to official status, and states the implicit rules: intent on every test, spec section linked per attribute file, `t.Run` for named table cases, and what the load helpers silently skip. AGENTS.md points coding agents at it.
- **`loadsAs(t, input, canonical)`** asserts an input compose YAML loads into the canonical model *expressed as YAML*: expectations read as compose documentation instead of Go struct literals. Converted `entrypoint` and `extra_hosts` tests as demonstrators.
- **Spec links**: every `loader/tests/` file now opens with the URL of the spec section it locks and a quote of the normative sentence, making the suite a navigable conformance matrix in both directions.
- **Completeness test**: every service attribute of the JSON schema must declare the test file locking it, or an explicit known-gaps entry — adding a schema property without deciding where it is tested now fails the suite. Tracked gaps: `mem_*`, `pull_refresh_after`, `read_only`, `shm_size`.
- **Hygiene**: named table cases now run as `t.Run` subtests, so a failing row identifies itself.

No behavior change; test-only (plus docs).


glours (approved):
LGTM

PR #915, opened 2026-08-17
fix(override): prevent panic on non-string list elements

glours:
`convertIntoMapping` performed unchecked type assertions when converting sequences (`depends_on`, `networks`, `models`) to mappings. Any non-string element, a YAML map ({}) or integer, caused a panic instead of a user-readable error.

Three panic paths are now closed:
- non-string element in a list: checked assertion returns an error
- scalar string shorthand (depends_on: web): handled as single-key map
- unrecognised type (e.g. integer scalar): explicit error instead of a silent nil that panicked downstream in `mergeMappings`

`mergeLogging` bare type assertions are guarded by the same approach. `mergeDependsOn`, `mergeModels` and `mergeNetworks` are consolidated into a shared `mergeAsMapping` helper.

### DIFF
--- a/override/merge.go
+++ b/override/merge.go
@@ -120,19 +120,25 @@ func mergeMappings(mapping map[string]any, other map[string]any, p tree.Path) (m
 }
 
 // logging driver options are merged only when both compose file define the same driver
-func mergeLogging(c any, o any, p tree.Path) (any, error) {
-	config := c.(map[string]any)
-	other := o.(map[string]any)
-	// we override logging config if source and override have the same driver set, or none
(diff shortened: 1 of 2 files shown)


maxproske:
It's not panicable now! :) 

```sh
cat > base.yaml <<'EOF'
services: {x: {image: alpine, depends_on: [y]}, y: {image: alpine}}
EOF
cat > over.yaml <<'EOF'
services: {x: {depends_on: [{}]}}
EOF

./compose-spec base.yaml over.yaml
# failed to load project: services.x.depends_on: cannot use map[string]interface {} as a mapping key
```

PR #921, opened 2026-08-25
replace uses of testify and update logrus to reduce indirect dependencies

thaJeztah:
### replace uses of testify

Most tests were using gotest.tools, so rewriting the ones that used
testify to use gotest.tools;

    git grep '"gotest.tools/v3' | wc -l
    129
    git grep '"github.com/stretchr/testify' | wc -l
    7

### chore(deps): github.com/sirupsen/logrus v1.10.1

update logrus to reduce transitive dependencies


thaJeztah:
cc @glours @ndeloof ptal

.golangci.yml
glours: We should also remove this block

glours (commented):
We should also remove the `testifylint` configuration from the `.golangci.yml` file

.golangci.yml
thaJeztah: Oh! Good catch; missed that one; fixed 👍

glours (approved):
LGTM

PR #922, opened 2026-08-25
introduce Project.WithoutUnresolvedOptionalDependencies

ndeloof:
Prunes optional (`required: false`) `depends_on` references to services absent from the model — disabled by an inactive profile or dropped by service selection. Required references are deliberately kept so consumers can detect them and report a meaningful error.

This cannot be folded into `WithProfiles`: `WithServicesEnabled` re-derives the enabled set by re-running `WithProfiles` over all services, and relies on dangling references surviving disablement so the dependency edge is restored when a profile is activated later. Hence an explicit final step, for consumers to apply once their selection is settled.

First consumer: docker/compose stops pruning these references as a `NewGraph` side effect that mutated the project (docker/compose#14081, Lot 0 step 4).

### DIFF
--- a/types/project.go
+++ b/types/project.go
@@ -551,6 +551,24 @@ func (p *Project) WithSelectedServices(names []string, options ...DependencyOpti
 	return newProject, nil
 }
 
+// WithoutUnresolvedOptionalDependencies removes from services any optional (required: false)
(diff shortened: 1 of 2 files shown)


glours (approved):
LGTM

PR #924, opened 2026-08-31
override: sequence merge ignores strictly-identical duplicate entries

ndeloof:
Merging sequences appended blindly: the same `env_file` declared in base + override was applied twice, an identical mount or device entry was duplicated and rejected by the engine at create time. Two strictly identical entries never carry more meaning than one.

Both merge regimes now drop an override entry deeply equal to one already present — the registered sequence merges and the default list merge — which folds `mergeExtraHosts` (the one merger that already deduplicated) into `mergeToSequence`. Entries that differ in form (short vs long syntax of the same declaration) are **not** equal and are kept: the rule is strict identity, never equivalence.

Property pinned by test: merging a value over an already-merged result is idempotent — the guarantee needed to resolve `pre_start` inheritance at load time (compose-spec/compose-go#866 follow-up) without breaking `config` round-trips.

### DIFF
--- a/override/merge.go
+++ b/override/merge.go
@@ -19,6 +19,7 @@ package override
 import (
 	"cmp"
 	"fmt"
+	"reflect"
 	"slices"
 
 	"github.com/compose-spec/compose-go/v2/tree"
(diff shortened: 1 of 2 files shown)


glours (approved):
Sounds good to me 👍

PR #925, opened 2026-09-02
fix: ignore default .env probe on permission denied

### LINKED ISSUE #14166: [BUG] permission denied on .env file in current directory lead to error
### Description

my current use case is using a sudo command to execute a tool that execute some docker commands...

to simplify the problem, running the following command fail with a permission denied (and btw the .env file did not exists)
```
[root@myhost ~]# sudo -u myuser docker compose -p myproject ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
stat /root/.env: permission denied
[root@myhost ~]# sudo -u myuser docker compose ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
stat /root/.env: permission denied
```
With current working directory changes it work as expected (no .env file in the /tmp directory)
```
[root@myhost ~]# cd /tmp
[root@myhost tmp]# sudo -u myuser docker compose ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
no configuration file provided: not found
[root@myhost tmp]# sudo -u myuser docker compose -p myproject ps -a --format "{{.ID}}|{{title .Status}}|{{.Ports}}"
29ddbf36662e|Exited (0) 4 Minutes Ago|
05ed8afdf90d|Exited (0) 4 Minutes Ago|
797d87bcc51e|Exited (0) 4 Minutes Ago|
6be65b4385c5|Exited (0) 4 Minutes Ago|
```

Why do not just print a warning for this permission denied ?
Is there an acceptable workarround for this ?

Thanks fj

### Compose Version

```Text
docker-compose-plugin-5.5.0-1.el9.x86_64
```

### Docker Environment

```Text
RedHat 9
container-selinux-2.235.0-1.el9_6.noarch
docker-ce-cli-29.7.2-1.el9.x86_64
containerd.io-2.3.4-1.el9.x86_64
docker-ce-rootless-extras-29.7.2-1.el9.x86_64
docker-ce-29.7.2-1.el9.x86_64
```

### Anything else?

i didn't test in different system, i guess it working like this

ndeloof:
The default `.env` lookup is an implicit convenience, not something the user asked for. When the working directory denies even probing for it (EACCES on stat — e.g. `sudo -u other-user` with a 0700 home as cwd), warn and skip the default instead of failing the whole command.

Conservative scope: only `os.IsPermission` is softened, and only for the implicit default — any other stat failure stays fatal, and an explicit env file behind the same permission wall still errors in `WithDotEnv`.

Fixes docker/compose#14166

### DIFF
--- a/cli/options.go
+++ b/cli/options.go
@@ -286,6 +286,15 @@ func WithEnvFiles(file ...string) ProjectOptionsFn {
 		if os.IsNotExist(err) {
 			return nil
 		}
+		if os.IsPermission(err) {
+			// The default .env is an implicit convenience, not something the
+			// user asked for: when permissions prevent even probing for it
+			// (say, running with a working directory owned by another user),
+			// warn and move on instead of failing the whole command. An
+			// explicit --env-file still fails hard in WithDotEnv.
+			logrus.Warnf("cannot access %s, ignoring default env file: %v", defaultDotEnv, err)
+			return nil
+		}
 		if err != nil {
 			return err
 		}
--- a/cli/options_test.go
+++ b/cli/options_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/compose-spec/compose-go/v2/types"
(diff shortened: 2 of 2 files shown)


thaJeztah:
I'm not 100% sure if this is correct though; running with a different user, then skipping `.env` could mean "now you're deploying to a different environment (e.g. production instead of test).

Given that there's already an escape hatch through the env-var, I'm not sure if we should do this.

thaJeztah:
(and the correct mitigation would be to `chmod` the `.env` file if it must be used, or `COMPOSE_DISABLE_ENV_FILE=1` if it's ok to ignore the `.env` file.

glours (approved):
LGTM

ndeloof:
> I'm not 100% sure if this is correct though; running with a different user, then skipping .env could mean "now you're deploying to a different environment (e.g. production instead of test).

This is also true if you run from the wrong folder, or any comparable situation where the target environment is set by `.env` and user unfortunately shoot into its own foot as compose loads a default .env file from PWD. Also consider compose looking for a compose.yaml file from parent folder(s). There are actually many ways to break a compose deployment if you're not careful, that's the price to pay for (too much?) flexibility introduced over time

PR #926, opened 2026-09-08
loader: screen the model against runtime-declared supported attributes

ndeloof:
The Compose Specification makes the runtime responsible for warning about attributes it parses but ignores (docker/compose#13150). Instead of each runtime hand-enumerating what it does **not** implement, the loader can screen the merged model against a declaration of what it **does**:

- `Options.SupportedAttributes` takes a set of `tree.Path` patterns; every attribute of the canonical, pre-normalization model (what the user actually wrote) matching none of them is reported through `Options.UnsupportedAttribute` (default: a logrus warning). Extension keys (`x-*`) are never reported; an undeclared subtree is reported once at its root.
- `tree.Matcher` gives `tree.Path` pattern sets a first-class matching API: exact `Matches` — declaring a path accepts that node only, so removing one leaf from a declared set reliably surfaces it — plus `MayContain` for walkers that must know when a declared attribute lives deeper.
- `schema.AttributePaths` derives the complete attribute-path inventory from the embedded JSON schema (pattern-keyed mappings as `*`, sequence items as `[]`, `$ref` cycles terminated, `^x-` pattern properties excluded — recorded as wildcards they would blanket-accept every undeclared sibling). A runtime builds its declaration by removing what it does not implement from the full specification, so every newly-specified attribute stays reported until deliberately wired in.

Example — a runtime declaring `schema.AttributePaths()` minus the swarm-only subtrees, loading the docker/compose#13149 file:

```
WARN compose.yaml: attribute services.app.deploy.update_config is not supported and will be ignored
WARN compose.yaml: attribute services.app.configs.[].uid is not supported and will be ignored
```


ndeloof:
will be combined with https://github.com/compose-spec/compose-go/pull/927

ndeloof:
Superseded by #927, merged as cda18529a with both halves combined: the deny patterns with value predicates from the original #927 plus the schema-derived fail-closed allowlist from this PR (landed into it via glours/compose-go#2). Same walk, same batch report.

PR #929, opened 2026-09-15
fix: type=image volume source is always a docker image reference

ndeloof:
This PR removes an undocumented, unintended behavior: a `type: image` volume's `source` silently stops being resolved as an image whenever it happens to match the name of another service in the project.

## Context

`05-services.md` documents `source` for a `type: image` mount as "a docker image reference" — nothing else. But `WithImagesResolved` currently special-cases it: if `source` matches an active or profile-disabled service name, digest resolution is skipped entirely and the raw name is passed through untouched, as if it were an implicit reference to that other service's locally built image.

This special-casing was never proposed or reviewed as a feature. It landed as a side effect of #894/#899 (wiring up dependent-image resolution for `WithImagesResolved`), is absent from the spec text and JSON schema, and has no integration anywhere else in the dependency graph: `DependsOn`, service selection (`WithSelectedServices`) and build ordering all remain unaware of it. In practice this means the "referenced" service isn't reliably built/pulled before the mount is used, so the only working case today is coincidental (the target image already exists locally under that name).

## What this PR brings

`type: image` volume sources are now always resolved as plain docker image references, consistently with the documented spec and with how the service's own `image:` and `pre_start` hook images are already handled. A source that isn't a syntactically valid docker reference (e.g. one that happens to reuse a mixed-case service name) now fails resolution with a clear error instead of being silently passed through.

If mounting another service's build output is wanted as an actual feature, it deserves an explicit, documented reference form — e.g. `source: service:xxx` — consistent with how the spec already expresses inter-service references elsewhere (`build.additional_contexts: [service:xxx]`), rather than an implicit fallback triggered by a name collision.

Found while reviewing docker/compose#13811.


glours (approved):
LGTM

PR #930, opened 2026-09-16
fix: an explicit working dir must survive a remote resource loader

### LINKED ISSUE #14224: When using docker compose OCI artifact, parameter --project-directory does not work for resolving relative volume paths
### Description

I have the following [docker compose artifact](https://hub.docker.com/r/usdotfhwaops/v2xhub-application) I am attempting to run with this command.
```
docker compose \
               -f "oci://docker.io/usdotfhwaops/v2xhub-application:develop" \
               --env-file "/home/dev/cdasim-scenario-runner/tmp/.env.street_1" \
               -p "street_1" --project-directory "/home/dev/cdasim-scenario-runner/tmp/" \
               up -d --yes --force-recreate 
```

Under the /home/dev/cdasim-scenario-runner/tmp directory I have local configuration files that are intended to get volumed into my services using relative paths. One example is [here ](https://github.com/usdot-fhwa-OPS/V2X-Hub/blob/66fa36b3beaa6da4fc9695d14468f6582e475526/configuration/docker-compose.yml#L20) where I volume a .sql file into my mysql service to initialize my database from a relative path.

When running this however and inspecting the containers I see the following resolution for this volume path
```

       "Mounts": [
            {
                "Type": "bind",
                "Source": "/home/dev/.cache/docker-compose/f22c7d5e9d784c051618dd5ae2b75a2c8132cefffe0c9aaed666a0ae8f1ab702/v2xhub_db_init_street_1",
                "Destination": "/docker-entrypoint-initdb.d",
                "Mode": "rw",
                "RW": true,
                "Propagation": "rprivate"
            },
```

This --project-directory parameter works fine for local docker compose files and there is nothing in the documentation of the CLI parameter or the docker compose OCI artifacts that to me suggests that this should not work . Am I missing something?

### Steps To Reproduce

Run 
```
docker compose \
               -f "oci://docker.io/usdotfhwaops/v2xhub-application:develop" \
               --env-file "/home/dev/cdasim-scenario-runner/tmp/.env.street_1" \
               -p "street_1" --project-directory "/home/dev/cdasim-scenario-runner/tmp/" \
               up -d --yes --force-recreate 
```

### Compose Version

Docker Compose version v5.5.1

### Docker Environment

```Text
Client: Docker Engine - Community
 Version:           29.8.0
 API version:       1.56
 Go version:        go1.26.8
 Git commit:        88096ef
 Built:             Thu Sep  3 21:50:45 2026
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          29.8.0
  API version:      1.56 (minimum version 1.40)
  Go version:       go1.26.8
  Git commit:       3ce5872
  Built:            Thu Sep  3 21:50:45 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.3.5
  GitCommit:        1294c24a7da8e5a793ed378161673abe94118892
 runc:
  Version:          1.5.1
  GitCommit:        v1.5.1-0-g8f2685a4
 docker-init:
  Version:          0.19.0

```

### Anything else?

_No response_

ndeloof:
Fixes docker/compose#14224: `docker compose -f oci://... --project-directory DIR` resolved relative bind mounts against the downloaded artifact's own cache directory instead of `DIR`.

## Context

`LoadConfigFiles` defaults the project's working dir to a remote resource loader's (git, oci) downloaded copy directory whenever it's still empty — intentional, so a self-contained remote artifact (`extends`, bundled env files) resolves against itself. But the check only looked at whether a working dir had already been set, never at whether the caller's `workingDir` argument was itself an explicit request (`--project-directory`) or just a default (the current directory). Once that value reaches `LoadConfigFiles` as a plain string, the two are indistinguishable — so an explicit request was silently overridden the same way a mere default would be.

## What this PR brings

`Options` gains `workingDirExplicit` (set via `SetWorkingDirExplicit`, mirroring the existing `projectName`/`projectNameImperativelySet` pair), and `LoadConfigFiles`'s remote-default assignment now skips itself when that flag is set. `cli.ProjectOptions.ReadConfigFiles` sets it from `options.WorkingDir != ""` — exactly the condition `GetWorkingDir` already uses to decide whether an explicit working dir was requested.

The self-contained-artifact default is untouched for the common case (no explicit `--project-directory`): only an explicit request is now protected from being overridden.


glours (approved):
LGTM

